PR #1, opened 2021-04-30
fix: ci lint for mod registry and image

justadogistaken:
do golangci-lint on registry&image module,  and fixed issues occurred


image/default_image.go
```diff
@@ -309,7 +309,9 @@ func (d DefaultImageService) uploadLayers(repo string, layers []v1.Layer, blobs
 						errCh <- err
 						return err
 					}
-					defer utils.CleanFile(file)
+					defer func() {
+						_ = utils.CleanFile(file)
```
fanux, line 313: better make a warn log

fanux:
should not ignore errors! @justadogistaken

image/default_image.go
justadogistaken: agree on this

fanux (approved):
handle errors

utils/file.go
```diff
@@ -146,12 +146,20 @@ func CopySingleFile(src, dst string) (int64, error) {
...
-	err = file.Close()
-	return os.Remove(file.Name())
+	// the following operation won't failed regularly, if failed, log it
+	err := file.Close()
+	if err != nil {
+		logger.Warn(err)
+	}
+	err = os.Remove(file.Name())
+	if err != nil {
+		logger.Warn(err)
```
fanux, line 160: The recommended way is to return the error, caller decision write log or not

justadogistaken: I suppose os.Remove/Close ... won't return err regularly. So if error occurs, we log it.

image/util.go
```diff
@@ -90,7 +90,9 @@ func GetClusterFileFromImageManifest(imageName string) string {
 func GetClusterFileFromBaseImage(imageName string) string {
 	mountTarget, _ := utils.MkTmpdir()
 	mountUpper, _ := utils.MkTmpdir()
-	defer utils.CleanDirs(mountTarget, mountUpper)
+	defer func() {
+		_ = utils.CleanDirs(mountTarget, mountUpper)
```
fanux, line 94: warn log

PR #14, opened 2021-04-30
feature(main): set docker default registry address

cuisongliu:

set docker.io is registry default address

None


justadogistaken:
we now take dockerhub as default registry? @fanux

fanux:
> we now take dockerhub as default registry? @fanux

No, using ACR or github registry by default

PR #42, opened 2021-05-07
layer value across goroutine when pull/push, which cause some problem

justadogistaken:
this PR fix problem of pull/push operation, where value layer would accross goroutines, which will result in sharing same layer file among all layers. Eventually, push is successed, but pull will fail

### Describe how you did it
make layer to be local variable to fix this problem. Because progress func Action will run in goroutines, meanwhile Action relys on layer.

### DIFF
--- a/image/default_image.go
+++ b/image/default_image.go
@@ -169,66 +169,73 @@ func (d DefaultImageService) downloadLayers(named reference.Named, manifest sche
 	errorCh := make(chan error, 2*len(manifest.Layers))
 	defer func() {
 		close(errorCh)
-		lerr := errors.New("failed to upload layers")
+		lerr := errors.New("failed to download layers")
 		for e := range errorCh {
 			err = errors.Wrap(e, lerr.Error())
 			lerr = err
 		}
 	}()
 
-	for _, layer := range manifest.Layers {
-		hex := layer.Digest.Hex()
-		shortHex := hex
-		if len(shortHex) > 12 {
-			shortHex = shortHex[0:12]
-		}
+	for _, lyr := range manifest.Layers {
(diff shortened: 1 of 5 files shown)


image/default_image.go
```diff
@@ -169,66 +169,73 @@ func (d DefaultImageService) downloadLayers(named reference.Named, manifest sche
...
-
-		// get layers stream first
-		blobReader, err := d.registry.DownloadLayer(context.Background(), named.Repo(), layer.Digest)
-		if err != nil {
-			flow.ShowMessage(shortHex+fmt.Sprintf(" failed to pull layer, err: %s", err), nil)
-			errorCh <- err
-			continue
-		}
+	for _, lyr := range manifest.Layers {
+		func(layer distribution.Descriptor) {
```
fanux, line 180: why you need this closure?

justadogistaken: for making layer a local valiable, cuz Action will be executing in goroutines

utils/compress/compress.go
```diff
@@ -133,7 +135,15 @@ func Dir(dir, target string) (err error) {
 
 // this uncompress will not change the metadata of original files
 func Uncompress(src io.Reader, dst string) error {
-	err := os.MkdirAll(dst, common.FileMode0755)
+	// need to set umask to be 000 for current process.
+	// there will be some files having higher permission like 777,
+	// eventually permission will be set to 755 when umask is 022.
+	_, err := system.Umask(0)
+	if err != nil {
+		return err
```
kakaZhou719, line 143: link to issue #49

kakaZhou719 (approved):

PR #47, opened 2021-05-07
feat(imp #46): replace all seadent to sealer

ysicing:

replace all seadent to sealer #46


build/testing/Clusterfile
```diff
@@ -3,7 +3,7 @@ kind: Cluster
 metadata:
   name: my-cluster
 spec:
-  image: seadent/my-kuberentes-cluster-with-dashboard:latest
+  image: sealer/my-kuberentes-cluster-with-dashboard:latest
```
fanux, line 6: the ACR sealer namespaces is not belong to us. using sealer-io instead

cert/kube_certs.go
```diff
@@ -153,7 +153,7 @@ func certList(CertPath, CertEtcdPath string) []Config {
 	}
 }
 
-type SeadentCertMetaData struct {
+type sealerCertMetaData struct {
```
fanux, line 156: using sealer to name a struct is not pretty good, using CertMetaData(may be lint failed) or MetaData instead

build/testing/Clusterfile
ysicing: ok

fanux (approved):

PR #59, opened 2021-05-10
fix scp error when create remote dir

kakaZhou719:
scp return error: "sftp: "Failure" (SSH_FX_FAILURE)" when we copy lcoal dir to remote

### Describe how you did it
replace "sftpClient.Mkdir:" to "sftpClient.MkdirAll:" 

comment of func sftpClient.Mkdir: An error will be returned if a file or  directory with the specified path already exists, or if the directory's parent folder does not exist
 

### Describe how to verify it
run build or apply cmd


oldthreefeng:
LGTM. if rebased.

Issue #24, opened 2021-05-04
sealer build many times at same cluster image


cuisongliu:
#### What happened:

build success , but can't push cluster image.  try again build.
<img width="1437" alt="image" src="https://user-images.githubusercontent.com/4150905/116995627-e1890c00-ad0c-11eb-8dc1-2270d48c55d4.png">

cd /var/lib/seadent/metadata
ls -l 

<img width="1124" alt="image" src="https://user-images.githubusercontent.com/4150905/116995718-01203480-ad0d-11eb-9923-0cc3b5c9e24e.png">

#### What you expected to happen:

I think it build one image for same content.
#### Environment:
- sealer version (use `sealer version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

kakaZhou719:
@cuisongliu 
1,push failed : you need run sealer login to set your docker config json. 
2, build one image yaml file at a time?

fanux:
Same file should has same layer hash to solve this issue I think. @justadogistaken

PR #62, opened 2021-05-10
Set image id as unique one

kakaZhou719:
### Describe how you did it
link to #24
set image ID as unique one

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -3,8 +3,9 @@ package parser
 import (
 	"bufio"
 	"fmt"
-	"github.com/alibaba/sealer/utils"
 	"strings"
+
+	"github.com/alibaba/sealer/utils"
 
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -3,6 +3,7 @@ package parser
 import (
 	"bufio"
 	"fmt"
+	"github.com/alibaba/sealer/utils"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


parser/parser.go
```diff
@@ -38,14 +39,14 @@ func (p *Parser) Parse(kubeFile []byte, name string) *v1.Image {
...
-
-		//TODO count layer hash
 		image.Spec.Layers = append(image.Spec.Layers, v1.Layer{
 			Hash:  "",
 			Type:  layerType,
 			Value: layerValue,
 		})
 	}
+	// gen consistent image ID
+	image.Spec.ID = utils.MD5(kubeFile, []byte(name))
```
justadogistaken, line 49: curious why generate id base on kubefile and image name instead of utils.GenUniqueID(32)

fanux: This is not appropriate and may even lead to bugs if two different image using same Clusterfile. Image ID should not using Clusterfile MD5

oldthreefeng: like rootfs image. it can be affect only the kubernetes version. not the others things.

oldthreefeng (commented):
so we must find the affect things of each image layer.

parser/parser.go
kakaZhou719: image id = md5sum(clusterfile+image name), same clusterfile with same image name guess that is a same image . what do you think ?@all

fanux: Not good..

justadogistaken: docker generate image id refer to content of image config. Is it fine to calculate image id depends on image metadata?

PR #66, opened 2021-05-11
feature: push image by name or id

tingxuan233:
       feature issue#37


fanux:
Docker is not supporting push an image using image ID, so we don't provide this feature too.

PR #84, opened 2021-05-13
fix not login and run a cloudImage will failed.

bxy4543:


cuisongliu:
please rebase it @bxy4543

PR #88, opened 2021-05-13
fix not login and run a cloudImage will failed

bxy4543:
fix not login and run a cloudImage will failed

### DIFF
--- a/runtime/utils.go
+++ b/runtime/utils.go
@@ -64,10 +64,12 @@ func PreInitMaster0(sshClient ssh.Interface, remoteHostIP string) error {
 	logger.Info("send cluster file to %s success !", remoteHostIP)
 
 	// send register login info
-	err = sshClient.Copy(remoteHostIP, common.DefaultRegistryAuthConfigDir(), common.DefaultRegistryAuthConfigDir())
-	if err != nil {
-		logger.Warn("send register config to remote host %s failed:%v", remoteHostIP, err)
-	} else {
+	authFile := common.DefaultRegistryAuthConfigDir()
+	if utils.IsFileExist(authFile) {
+		err = sshClient.Copy(remoteHostIP, authFile, authFile)
+		if err != nil {
+			return fmt.Errorf("failed to send register config %s to remote host %s err: %v", authFile, remoteHostIP, err)
+		}
 		logger.Info("send register info to %s success !", remoteHostIP)
 	}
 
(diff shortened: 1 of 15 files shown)


runtime/utils.go
```diff
@@ -62,11 +63,14 @@ func PreInitMaster0(sshClient ssh.Interface, remoteHostIP string) error {
...
+	authFile := common.DefaultRegistryAuthConfigDir()
+	//if auth file exist, copy file to remote
+	if _, err := os.Stat(authFile); err == nil {
+		// send register login info
+		err = sshClient.Copy(remoteHostIP, authFile, authFile)
+		if err != nil {
+			return fmt.Errorf("send register config to remote host %s failed:%v", remoteHostIP, err)
+		}
 	}
 	logger.Info("send register info to %s success !", remoteHostIP)
```
justadogistaken, line 75: what is the solution if authFile not exists? still log success?

justadogistaken: send registry auth info , not register info.
and the success log would still be executed, I it's not allowed

runtime/utils.go
```diff
@@ -66,7 +66,7 @@ func PreInitMaster0(sshClient ssh.Interface, remoteHostIP string) error {
 	// send register login info
 	err = sshClient.Copy(remoteHostIP, common.DefaultRegistryAuthConfigDir(), common.DefaultRegistryAuthConfigDir())
```
justadogistaken, line 67: we'd better check file existence first

runtime/utils.go
```diff
@@ -64,11 +64,14 @@ func PreInitMaster0(sshClient ssh.Interface, remoteHostIP string) error {
...
-	err = sshClient.Copy(remoteHostIP, common.DefaultRegistryAuthConfigDir(), common.DefaultRegistryAuthConfigDir())
-	if err != nil {
-		return fmt.Errorf("send register config to remote host %s failed:%v", remoteHostIP, err)
+	authFile := common.DefaultRegistryAuthConfigDir()
+	if utils.IsFileExist(authFile) {
+		err = sshClient.Copy(remoteHostIP, authFile, authFile)
+		if err != nil {
+			return fmt.Errorf("failed to send register config %s to remote host %s err: %v", authFile, remoteHostIP, err)
+		}
+		logger.Info("send register info to %s success !", remoteHostIP)
```
justadogistaken, line 73: we should ensure whether authFile is necessary, and if we should warn users when the file does not exist

fanux:
LGTM  @justadogistaken

fanux (approved):

justadogistaken (approved):

PR #100, opened 2021-05-14
Fix/run one master cluster

bxy4543:

### DIFF
--- a/apply/run.go
+++ b/apply/run.go
@@ -14,6 +14,7 @@ import (
 
 type ClusterArgs struct {
 	cluster    *v1.Cluster
+	imageName  string
 	nodeArgs   string
 	masterArgs string
 	user       string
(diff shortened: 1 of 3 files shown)


apply/run.go
```diff
@@ -40,22 +41,27 @@ func IsIPList(args string) bool {
...
-		c.cluster.Spec.SSH.Passwd = c.passwd
-		c.cluster.Spec.Provider = common.AliCloud
-		return nil
+	c.cluster.Spec.Image = c.imageName
+	if IsNumber(c.masterArgs) {
+		if IsNumber(c.nodeArgs) || c.nodeArgs == "" {
+			c.cluster.Spec.Masters.Count = c.masterArgs
+			c.cluster.Spec.Nodes.Count = c.nodeArgs
+			c.cluster.Spec.SSH.Passwd = c.passwd
+			c.cluster.Spec.Provider = common.AliCloud
```
oldthreefeng, line 50: c.cluster.Spec.Provider need to read from Cluster.  or set a default CloudProvider
If user want to Use other Cloud . we may suport for multi CloudProvider soon?  

```
DefaultCloudProvider = AliCloud
common.DefaultCloudProvider
```

oldthreefeng (commented):
LGTM. just Run a single master cluster will couse the error.

apply/run.go
cuisongliu: or add cloud provider , if masterArgs is number then replace the provider

apply/run.go
```diff
@@ -14,6 +14,7 @@ import (
 
 type ClusterArgs struct {
 	cluster    *v1.Cluster
+	imageName  string
```
cuisongliu, line 17: this args should be from cmd/run.go transfer to here. I not found `imageName` args in cmd/run command, you not finished yet?

cuisongliu (dismissed):
please add images args in cmd/run.go

apply/run.go
```diff
@@ -40,22 +41,27 @@ func IsIPList(args string) bool {
...
 
 func (c *ClusterArgs) SetClusterArgs() error {
-	if IsNumber(c.masterArgs) && IsNumber(c.nodeArgs) {
-		c.cluster.Spec.Masters.Count = c.masterArgs
-		c.cluster.Spec.Nodes.Count = c.nodeArgs
-		c.cluster.Spec.SSH.Passwd = c.passwd
-		c.cluster.Spec.Provider = common.AliCloud
-		return nil
+	c.cluster.Spec.Image = c.imageName
+	if IsNumber(c.masterArgs) {
```
fanux, line 45: if isnumber(master) && (isNUmber(node) || nodeargs == "")

fanux (approved):

PR #136, opened 2021-05-20
feat(main): cni feature #81 no.5

cuisongliu:

### DIFF
--- a/apply/default_applier.go
+++ b/apply/default_applier.go
@@ -173,7 +173,8 @@ func (c *DefaultApplier) diff() (todoList []ActionName, err error) {
 
 	// if only contains PullIfNotExist and Mount, we do nothing
 	if len(todoList) == 2 {
-		return nil, nil
+		todoList = append(todoList, CNI)
+		return todoList, nil
 	}
 
 	todoList = append(todoList, CNI)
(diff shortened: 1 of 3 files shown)


apply/default_applier.go
```diff
@@ -74,6 +75,9 @@ var ActionFuncMap = map[ActionName]func(*DefaultApplier) error{
 	Guest: func(applier *DefaultApplier) error {
 		return applier.Guest.Apply(applier.ClusterDesired)
 	},
+	CNI: func(applier *DefaultApplier) error {
```
fanux, line 78: I think applier not care about CNI in the feature

runtime/runtime.go
```diff
@@ -14,6 +14,7 @@ type Interface interface {
 	Hook(cluster *v1.Cluster) error
 	Upgrade(cluster *v1.Cluster) error
 	Reset(cluster *v1.Cluster) error
+	CNI(cluster *v1.Cluster) error
```
fanux, line 17: delete this interface, just apply it after master0 init

apply/default_applier.go
cuisongliu: so , we not support only change cni configure ?

fanux (approved):

PR #160, opened 2021-05-24
feat(main): fix root and metadata

cuisongliu:


runtime/init.go
```diff
@@ -106,7 +102,18 @@ func (d *Default) initRunner(cluster *v1.Cluster) error {
 
 	return nil
 }
-
+func (d *Default) CopyMetadata() error {
```
fanux, line 105: Why we need copy Metadata?

cuisongliu: at init,applyNodes,applyMasters process need Metadata,  but Metadata in rootfs   and read metadata not mount /tmp/<cluster-name> yet , read metada at runtime init func so far.

fanux: Using /var/lib/sealer/../Metadata

cuisongliu: yes, i using it .but in runtime init this dir not exists  yet

fanux: Newruntime after Mount image, so that you needn't copy metadata

fanux (approved):

PR #195, opened 2021-05-26
feat: uncouple between layer-id and descriptor-hash

justadogistaken:
After this pr committed, there have some changes under layer storage. We add three files to record more info for supporting more complex scenes. There are some introductions for new files following:
size: size of layer data
distribution_layer_digest: created after pushed or pull a image. It will record the digest for layer in specific repository. With this, it will avoid pushing image duplicately, but if the file does not exist, sealer will push layer anyway.
tar-data.json.gz: it is used to recover a layer data, as long as the layer file content(not include header info) stay consistent, we could recover the entire layer data with this. 
未命名文件


touchuyht (approved):

touchuyht (dismissed):

fanux (approved):

fanux:
LGTM /need tests

kakaZhou719:
could we pls  split this PR(51 files) as more ? its hard to me to read it. and do rebase ,this file（testhelper/k8s_client.go）is no longer in use。

justadogistaken:
> could we pls split this PR(51 files) as more ? its hard for me to read it. and do rebase ,this file（testhelper/k8s_client.go）is no longer in use。

I don't see why you want me to divide the pr to multiple prs. The pr aims to implement a new single feature, it's also weird and hard for me to split the pr. And many changes are vendor, you could ignore those changes actually.

image/store/distribution.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return err
+	}
+	// override metadata items, and add new metadata
+	for key, value := range newMetadatas {
+		metadataMap[key] = value
+	}
+
+	distributionMetadatas := DistributionMetadata{}
+	for key, value := range newMetadatas {
```
touchuyht, line 72: Is it a typing error that newMetadatas instead of metadataMap appears after range?

justadogistaken: Yes, I was wrong, it should be metadataMap.

PR #198, opened 2021-05-27
fix sealer apply repeatedly copying remote existing files

### LINKED ISSUE #164: Send sealer  binary file  repeatedly
#### What happened:
1,run sealer apply -f (my raw clusterfile,will apply infra)
2, shrink cluster node , i run sealer apply -f ($HOME/.sealer/Clusterfile,remote infra and cluster context already exist), it still send  sealer binary file . cost more time 

image

no need to send sealer  binary file  repeatedly, because sealer  binary file is already exist
#### Environment:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

touchuyht:
Sealer apply will copy the files to remote host even if that file already exists because there is no checking process before copy. The new copy function will first check if the remote file exists and if it exists, it will check the md5sum of that remote file and compare it with local file's md5sum value. The copying process only happens when remote file does not exist or is stale.

Fixes #164

### DIFF
--- a/utils/ssh/scp.go
+++ b/utils/ssh/scp.go
@@ -217,8 +217,20 @@ func (s *SSH) copyLocalDirToRemote(host string, sshClient *ssh.Client, sftpClien
 	}
 }
 
+// check the remote file existence before copying
 // solve the sesion
 func (s *SSH) copyLocalFileToRemote(host string, sshClient *ssh.Client, sftpClient *sftp.Client, localPath, remotePath string) error {
+	var (
+		srcMd5, dstMd5 string
+	)
+	srcMd5 = LocalMd5Sum(localPath)
+	if s.IsFileExist(host, remotePath) {
+		dstMd5 = s.RemoteMd5Sum(host, remotePath)
+		if srcMd5 == dstMd5 {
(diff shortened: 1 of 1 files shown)


kakaZhou719 (approved):
LGTM

utils/ssh/scp.go
```diff
@@ -217,8 +217,20 @@ func (s *SSH) copyLocalDirToRemote(host string, sshClient *ssh.Client, sftpClien
 	}
 }
 
+// check the remote file existence before copying
 // solve the sesion
 func (s *SSH) copyLocalFileToRemote(host string, sshClient *ssh.Client, sftpClient *sftp.Client, localPath, remotePath string) error {
+	var (
+		srcMd5, dstMd5 string
+	)
+	srcMd5 = FromLocal(localPath)
```
fanux, line 226: I think we can rename this function.

fanux: like GetLocalFileMD5

justadogistaken (approved):
LGTM

utils/ssh/scp.go
touchuyht: solved

PR #200, opened 2021-05-27
fix scale down nodes ClOUD mode

fanux:


apply/cloud_applier.go
```diff
@@ -35,13 +35,42 @@ func NewAliCloudProvider(cluster *v1.Cluster) Interface {
 	return &CloudApplier{d}
 }
 
+func (c *CloudApplier) ScaleDownNodes(cluster *v1.Cluster) (isScaleDown bool, err error) {
```
justadogistaken, line 38: This is a ScaleDownNodes func, but it update c.NodesToJoin within the func, feel confused

apply/current_cluster.go
```diff
@@ -45,6 +54,8 @@ func getCurrentNodes() (*v1.Cluster, error) {
 		}
 		cluster.Spec.Nodes.IPList = append(cluster.Spec.Nodes.IPList, addr)
 	}
```
justadogistaken, line 56: Feel curious about "Count", why we need to keep this value? just len(...) is fine?

apply/cloud_applier.go
```diff
@@ -35,13 +35,42 @@ func NewAliCloudProvider(cluster *v1.Cluster) Interface {
...
+	logger.Info("desired master %s, current master %s, desired nodes %s, current nodes %s", c.ClusterDesired.Spec.Masters.Count,
+		cluster.Spec.Masters.Count,
+		c.ClusterDesired.Spec.Nodes.Count,
+		cluster.Spec.Nodes.Count)
+	if c.ClusterDesired.Spec.Masters.Count >= cluster.Spec.Masters.Count &&
```
justadogistaken, line 43: the invoking chain is too long, would it be better to declare the values at the beginning? just a suggestion

apply/cloud_applier.go
fanux: Oh, we needn't update it, I will fixed it

apply/current_cluster.go
fanux: User will edit this value in CLOUD mod

apply/cloud_applier.go
fanux: I think it's OKay.

kakaZhou719 (approved):

apply/cloud_applier.go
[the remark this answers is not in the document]
fanux: infra and baremetal Apply will check it.

PR #209, opened 2021-05-27
fix hash of tar diff

### LINKED ISSUE #151: sealer push image override
#### What happened:
image

#### Environment:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

justadogistaken:
hash of tar contains many header info, so hash will change in different machine, so I hardcode the gname and uname for the file header to avoid this problem.

Fixes #151 
### Special notes for reviews
this is just a temp solution. A better layer arch is more needed.


touchuyht (approved):
LGTM

cuisongliu (approved):

PR #210, opened 2021-05-28
fix get instances ip bug

formy99:


infra/aliyun/ali_ecs.go
```diff
@@ -49,22 +49,29 @@ func (a *AliProvider) TryGetInstance(request *ecs.DescribeInstancesRequest, resp
 		}
 		instances := response.Instances.Instance
 		if expectCount != -1 {
```
fanux, line 51: if expectCount == -1 {
   return nil 
}

formy99: optimized

fanux (approved):

PR #222, opened 2021-05-28
feat: #182 auto detect network interface name by master ip

### LINKED ISSUE #182: auto detect network interface name by master ip.

jinnzy:

Fixes #182


cuisongliu:
if diff node use diff interface ?? I think interface name shloud be stay

PR #224, opened 2021-05-28
add copy kubectl and kubeconfig BAREMETAL

fanux:


runtime/utils.go
```diff
@@ -76,3 +79,25 @@ func PreInitMaster0(sshClient ssh.Interface, remoteHostIP string) error {
...
 	return nil
 }
+
+func GetKubectlAndKubeconfig(ssh ssh.Interface, host string) error {
+	// fetch the cluster kubeconfig, and add /etc/hosts "EIP apiserver.cluster.local" so we can get the current cluster status later
+	err := ssh.Fetch(host, path.Join(common.DefaultKubeConfigDir(), "config"), common.KubeAdminConf)
+	if err != nil {
+		return errors.Wrap(err, "failed to copy kubeconfig")
+	}
+	err = utils.AppendFile(common.EtcHosts, fmt.Sprintf("%s %s", host, common.APIServerDomain))
```
cuisongliu, line 89: need check etc host `common.APIServerDomain` is exist,some node this host config host is  localhost or vip

runtime/init.go
```diff
@@ -63,9 +63,20 @@ func (d *Default) init(cluster *v1.Cluster) error {
 		return err
 	}
 
+	if err := d.GetKubectlAndKubeconfig(); err != nil {
```
cuisongliu, line 66: if master is changed. only at init step is not complete. I think the copy config need at all node change step.

runtime/utils.go
fanux: AppendFile already check the content is already exist, but not check the APIServerDomain, I will fix it in other PR

runtime/init.go
fanux: If changed, User copy it by himself, If you copy it every time, it's inefficient.

cuisongliu (approved):

PR #233, opened 2021-06-01
feature: Add E2E test base file for sealer

kakaZhou719:
Adding E2e test for sealer apply to keep our code more robust

### Describe how you did it
1, split github work flow as more jobs
2, add test framework for sealer ,will run sealer as non-root user.
3, add utils file and fixtures 

### Describe how to verify it

install ginkgo and config your env as readme.md said.
ginkgo -v --focus="sealer apply" test


test/suites/build/build.go
```diff
@@ -1,22 +1,99 @@
...
+
+func GetTestClusterFilePath() string {
+	return ""
+}
+
+type ArgsOfBuild struct {
+	KubeFile, ImageName, Context, BuildType string
+}
+
+func (a *ArgsOfBuild) SetKubeFile(kubeFile string) *ArgsOfBuild {
```
justadogistaken, line 51: attribution could be set using "a.KubeFile = kubeFile", why encapsulate the operation?

test/suites/build/build.go
```diff
@@ -1,22 +1,99 @@
...
+	a.Context = context
+	return a
+}
+
+func (a *ArgsOfBuild) SetBuildType(buildType string) *ArgsOfBuild {
+	a.BuildType = buildType
+	return a
+}
+
+func (a *ArgsOfBuild) ToString() string {
```
justadogistaken, line 71: If you do the operations above for implementing ArgsOfBuild in Builder mode, "ToString()" to "Build()" would be better, or just String(). just a suggestion.

test/suites/build/build.go
```diff
@@ -1,22 +1,99 @@
...
+	return fmt.Sprintf("%s build -f %s -t %s -c %s -b %s", settings.DefaultSealerBin, a.KubeFile, a.ImageName, a.Context, a.BuildType)
```
justadogistaken, line 72: Can we guarantee those params are not empty? if not so, maybe validate them first would be nicer.

test/suites/build/build.go
```diff
@@ -1,22 +1,99 @@
...
+}
+
+func NewArgsOfBuild() *ArgsOfBuild {
+	return &ArgsOfBuild{}
```
justadogistaken, line 76: NewArgsOfBuild is too simple, I think let call do "&ArgsOfBuild{}" would be fine.

test/testhelper/utils.go
```diff
@@ -1,13 +1,82 @@
...
 }
+
+func CreateTempFile() string {
+	dir := os.TempDir()
+	file, err := ioutil.TempFile(dir, "tmpfile")
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	defer func() {
+		err := file.Close()
+		if err != nil {
+			fmt.Println(err)
```
justadogistaken, line 30: Is there any reason for fmt.Println(err) instead of logger?

test/suites/build/build.go
kakaZhou719: truely, seem like  "Build()"  more better than "ToString()"

test/suites/build/build.go
kakaZhou719: yes, need to do validate .

test/suites/build/build.go
fanux: I think that's okay, because you don't know the set logic will change or not in the future.

justadogistaken (approved):
LGTM if you've tested already

PR #234, opened 2021-06-01
fix apply cloud master

fanux:


fanux:
need using kubernetes:v1.19.9-1 to test this PR

cuisongliu (approved):
add node or master test is ok ? loadMetadata in add master or node is ok ?

PR #252, opened 2021-06-03
get cluster status after add localhost kubectl and kubeconfig

ly-wjj:
After the master is initialized, you can simply verify whether the installation of kubectl and master is ok

### Describe how you did it
I use the kubectl get nodes command for simple verification


runtime/init.go
```diff
@@ -310,3 +314,14 @@ func (d *Default) decodeJoinCmd(cmd string) {
 	}
 	logger.Debug("joinToken: %v\nTokenCaCertHash: %v\nCertificateKey: %v", d.JoinToken, d.TokenCaCertHash, d.CertificateKey)
 }
+
+// get nodes list,check the config of kubectl
+func (d *Default) getClusterNodes() error {
+	output, err := utils.CmdOutput("kubectl", "get", "nodes")
```
fanux, line 320: If we get nodes immediately, the state of the node is likely to be NotReady.

ly-wjj: I think this is just a staged verification.I can verify whether the kubectl‘s configuration is correct and whether the master starts normally. This can detect problems in advance.

fanux: I think we can take all the psychic in an individual command, like sealer check --post(--pre). If show the Not Ready status, user may be confused about it.

fanux:
provides this feature in post check, close it now 😃

Issue #263, opened 2021-06-04
unmount rootfs failed when delete Clusterfile 


bxy4543:
#### What happened:
sealer delete -f /root/.sealer/my-cluster/Clusterfile ：
```shell
/bin/sh: /var/lib/sealer/data/my-cluster/rootfs/scripts/clean.sh: No such file or directory
2021-06-04 15:53:54 [EROR] exec command failed Process exited with status 127
2021-06-04 15:53:54 [EROR] 172.16.0.165:exec /bin/sh -c /var/lib/sealer/data/my-cluster/rootfs/scripts/clean.sh failed, exec command failed 172.16.0.165 /bin/sh -c /var/lib/sealer/data/my-cluster/rootfs/scripts/clean.sh
2021-06-04 15:53:55 [INFO] [ssh][172.16.0.169] : rm -rf /var/lib/sealer/data/my-cluster/rootfs
2021-06-04 15:53:55 [INFO] [ssh][172.16.0.168] : rm -rf /var/lib/sealer/data/my-cluster/rootfs
2021-06-04 15:53:55 [INFO] [ssh][172.16.0.167] : rm -rf /var/lib/sealer/data/my-cluster/rootfs
2021-06-04 15:53:55 [INFO] [ssh][172.16.0.166] : rm -rf /var/lib/sealer/data/my-cluster/rootfs
2021-06-04 15:53:55 [EROR] unmountRootfs failed
```

#### What you expected to happen:

the execution machine is master0

sealer run registry.cn-qingdao.aliyuncs.com/sealer-io/kubernetes:v1.19.9.alpha.1 --masters 172.16.0.165,172.16.0.166,172.16.0.167 --nodes 172.16.0.168,172.16.0.169 --passwd ***

 then exec 

sealer delete -f /root/.sealer/my-cluster/Clusterfile 

rootfs already delete

clcc2019:
sh script/clean.sh

use umount xxpath

kakaZhou719:
![image](https://user-images.githubusercontent.com/83740799/121161133-450fd780-c87f-11eb-81a4-32389e9771db.png)

bxy4543:
#286

PR #286, opened 2021-06-08
Fix delete failure after master0 execution

bxy4543:
fix delete failure after master0 execution

#263 

### Describe how to verify it
The CleanFS method will do this

### DIFF
--- a/runtime/reset.go
+++ b/runtime/reset.go
@@ -33,7 +33,7 @@ func (d *Default) reset(cluster *v1.Cluster) error {
 	if err != nil {
 		logger.Error("failed to clean masters %v", err)
 	}
-	err = utils.CleanFiles(common.GetClusterWorkDir(cluster.Name), common.DefaultClusterBaseDir(cluster.Name), common.DefaultKubeConfigDir())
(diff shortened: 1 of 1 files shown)


runtime/reset.go
```diff
@@ -33,7 +33,7 @@ func (d *Default) reset(cluster *v1.Cluster) error {
 	if err != nil {
 		logger.Error("failed to clean masters %v", err)
 	}
-	err = utils.CleanFiles(common.GetClusterWorkDir(cluster.Name), common.DefaultClusterBaseDir(cluster.Name), common.DefaultKubeConfigDir())
+	err = utils.CleanFiles(common.GetClusterWorkDir(cluster.Name), common.DefaultKubeConfigDir())
```
fanux, line 36: So how to clean the cluster basedir(/var/lib/sealer/data/my-cluster)

bxy4543: this function will do this：
	}

fanux (approved):

kakaZhou719:
seems like this PR does not solve this problem.
image

Issue #108, opened 2021-05-18
validate auth info before building a image


justadogistaken:
#### What happened:
sealer build will push image after building done. And sealer build -t cloudfs:xxx will be push to "registry.cn-qingdao.aliyuncs.com" by default.
#### What you expected to happen:
pushing to "registry.cn-qingdao.aliyuncs.com" by default is weird. I think registry url should be specified when push a image.
#### Environment:
- sealer version (use `sealer version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

fanux:
Check user permission before builds, if user doesn't have push access, return error tells him to specify the URL

ly-wjj:
I will fix it

ly-wjj:
@fanux @kakaZhou719 #338 pull request has resolved the issue. Can this issue be closed?

PR #289, opened 2021-06-08
fix bug: validate auth info before building a image #108

ly-wjj:

### DIFF
--- a/image/distributionutil/registry.go
+++ b/image/distributionutil/registry.go
@@ -29,6 +29,7 @@ func NewV2Repository(named reference.Named, actions ...string) (distribution.Rep
 	authConfig, err := utils.GetDockerAuthInfoFromDocker(named.Domain())
 	if err != nil {
 		logger.Warn("failed to get auth info, err: %s", err)
+		return nil, err
 	}
 
 	repo, err := NewRepository(context.Background(), authConfig, named.Repo(), registryConfig{Insecure: true, Domain: named.Domain()}, actions...)


image/distributionutil/registry.go
```diff
@@ -29,6 +29,7 @@ func NewV2Repository(named reference.Named, actions ...string) (distribution.Rep
 	authConfig, err := utils.GetDockerAuthInfoFromDocker(named.Domain())
 	if err != nil {
 		logger.Warn("failed to get auth info, err: %s", err)
+		return nil, err
```
fanux, line 32: Sometimes we don't need login the registry when we build CloudImage if Kubefile only has COPY command. @kakaZhou719 double check this PR

image/distributionutil/registry.go
```diff
@@ -29,6 +29,7 @@ func NewV2Repository(named reference.Named, actions ...string) (distribution.Rep
 	authConfig, err := utils.GetDockerAuthInfoFromDocker(named.Domain())
 	if err != nil {
 		logger.Warn("failed to get auth info, err: %s", err)
```
justadogistaken, line 31: This is not the point.
sometimes we do not need to login to registry(like sealer pull).

fanux:
@justadogistaken but the build will failed if Kubefile has CMD or RUN command on push. Not a best practice. So we rellay need login on CLOUD build mod @kakaZhou719

kakaZhou719:
@justadogistaken ，sealer pull private image will be successful ?  if so , i think we do not need to login to registry except its cloud build scenario. because cloud build will push images to registry by default, docker.json will be sent to cloud before cloud build start.@fanux.

kakaZhou719:
@ly-wjj , as we discussed, the better way we fix this bug is , checking the docker.json file before cloud build. its necessary to do precheck for cloud build.

PR #311, opened 2021-06-11
local build wipe off copy file's  source path

carlos-lee:
local build fit this Kubefile：
FROM registry.cn-qingdao.aliyuncs.com/sealer-io/kubernetes:v1.19.9
COPY /Users/zcy/Documents/sealer-src/recommended.yaml .
CMD kubectl apply -f recommended.yaml

### Describe how you did it
wipe off copy file source path

### Describe how to verify it
local build success


fanux (dismissed):

build/local_builder.go
```diff
@@ -256,7 +257,7 @@ func (l *LocalBuilder) execLayer(layer *v1.Layer, tempTarget string) error {
 			dist = filepath.Join(tempTarget, strings.Fields(layer.Value)[1])
 		} else {
 			// src is file
-			dist = filepath.Join(tempTarget, strings.Fields(layer.Value)[1], strings.Fields(layer.Value)[0])
+			dist = filepath.Join(tempTarget, strings.Fields(layer.Value)[1], path.Base(strings.Fields(layer.Value)[0]))
```
kakaZhou719, line 260: you could replace ‘path’ with ‘filepath’ to ensure our code will be matched with actual file system.

kakaZhou719: by the way , you can comment your PR  /test build to run e2e test for build.

kakaZhou719, line 260: i think we should keep the same principle with docker COPY. that is mean:  the <src> path must be inside the context of the build. and the <dest> path must be an absolute path inside the container.

kakaZhou719 (commented):
It’s better not to modify this at present。

build/local_builder.go
carlos-lee: as you say,
```
Copy sth /etc/sth
```
in sealer context is  absolute path /etc/sth

carlos-lee: absolute path inside the container is not related  to Copy source path, is related to Copy destination path

build/local_builder.go
```diff
@@ -247,7 +248,7 @@ func (l *LocalBuilder) execLayer(layer *v1.Layer, tempTarget string) error {
 			dest = filepath.Join(tempTarget, strings.Fields(layer.Value)[1])
 		} else {
 			// src is file
-			dest = filepath.Join(tempTarget, strings.Fields(layer.Value)[1], strings.Fields(layer.Value)[0])
+			dest = filepath.Join(tempTarget, strings.Fields(layer.Value)[1], path.Base(strings.Fields(layer.Value)[0]))
```
justadogistaken, line 251: just do filepath.Base(), do not need to import another pkg.

justadogistaken:
@kakaZhou719 Do we have restriction about build context? Like disallowing that building image out of current dir.

PR #322, opened 2021-06-15
optimize dir busy umount image

### LINKED ISSUE #314: Clusterfile API upgrade.
We need to encapsulate access to the *v1.Cluster data.
   For example, get  master IPList:
   Now we directly access the cluster.spec.master.IPlist
   but if we upgrade the cluster struct
   we need to update a lot of codes.
   So provide a function to get it is more Reasonable.

```
func (c *v1.Cluster)GetMasterIPList()([]string error){
     return cluster.spec.master.IPList
}
```

bxy4543:

Fix umount failure affecting apply

Delete $home/.sealer/Cluster at the end

fix #314

### DIFF
--- a/filesystem/filesystem.go
+++ b/filesystem/filesystem.go
@@ -19,6 +19,9 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
+
+	infraUtils "github.com/alibaba/sealer/infra/utils"
 
 	"github.com/alibaba/sealer/utils"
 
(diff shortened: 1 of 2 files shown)


kakaZhou719:
does this PR linked issues ?pls add it.

filesystem/filesystem.go
```diff
@@ -59,18 +60,23 @@ func IsDir(path string) bool {
...
 }
 
 func (c *FileSystem) umountImage(cluster *v1.Cluster) error {
 	mountdir := common.DefaultMountCloudImageDir(cluster.Name)
 	if utils.IsFileExist(mountdir) {
-		logger.Debug("unmount cluster dir %s", mountdir)
-		if err := mount.NewMountDriver().Unmount(mountdir); err != nil {
-			logger.Warn("failed to unmount %s, err: %v", mountdir, err)
+		var err error
+		for i := 0; i < 10; i++ {
```
fanux, line 70: Maybe we can use utils.Retry function.

fanux (approved):

PR #335, opened 2021-06-18
optimize: do layer filesystem initialization on start

justadogistaken:
There is a new package named onstart, we can do some initialization job within the package.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -30,19 +30,11 @@ const (
 )
 
 const (
-	DefaultImageRootDir           = "/var/lib/sealer/data"
 	DefaultWorkDir                = "/var/lib/sealer/%s/workdir"
 	DefaultClusterFileName        = "Clusterfile"
-	DefaultMetadataName           = "Metadata"
 	DefaultClusterRootfsDir       = "/var/lib/sealer/data"
 	DefaultClusterInitFile        = "/var/lib/sealer/data/%s/scripts/init.sh"
 	DefaultClusterClearFile       = "/var/lib/sealer/data/%s/rootfs/scripts/clean.sh"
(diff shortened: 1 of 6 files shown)


sealer/onstart/onstart.go
```diff
@@ -0,0 +1,36 @@
+package onstart
```
fanux, line 1: The name of the module is a little strange.

fanux: How about call this module `boot`?

image/store/filesystem.go
```diff
@@ -172,6 +172,16 @@ func (ls LayerStorage) assembleTar(id LayerID) (io.ReadCloser, error) {
 	}), nil
 }
 
+func InitLayerDirectory() error {
+	dls := NewDefaultLayerStorage()
```
touchuyht, line 176: I wonder if it is necessary to call this function here. We could just use layerDataRoot and layerDBRoot.

sealer/onstart/onstart.go
```diff
@@ -0,0 +1,36 @@
...
+
+	err = utils.MkDirIfNotExists(common.DefaultImageMetaRootDir)
+	if err != nil {
+		return fmt.Errorf("failed to mk sealer metadata root dir, err: %s", err)
+	}
+	return nil
+}
+
+func OnBoot() error {
+	err := initRootDirectory()
```
touchuyht, line 26: The actions of theses functions are similar, maybe we could create an array which saves names of the files to be created and we range the array to create them.

justadogistaken:
@fanux @touchuyht PTAL

fanux (approved):

PR #342, opened 2021-06-22
Fix/clusterfile not exist

bxy4543:
optimize errors caused by clusterfile not found.

add Cluster check


apply/default_applier.go
```diff
@@ -120,8 +121,33 @@ var ActionFuncMap = map[ActionName]func(*DefaultApplier) error{
 	CleanFS: func(applier *DefaultApplier) error {
 		return applier.FileSystem.Clean(applier.ClusterDesired)
 	},
+	CheckDesiredCluster: func(applier *DefaultApplier) error {
+		return checkDesiredCluster(applier)
+	},
 }
 
+func checkDesiredCluster(applier *DefaultApplier) error {
```
fanux, line 129: I think we can add this function to the check module.

fanux: pre check

utils/hosts.go
```diff
@@ -38,3 +38,27 @@ func GetDiffHosts(hostsOld, hostsNew v1.Hosts) (add, sub []string) {
 
 	return
 }
+
+func HasSameHosts(hostsOld, hostsNew v1.Hosts) bool {
```
kakaZhou719, line 42: i think only IPlist we need in this function. (hostsOld, hostsNew []sting)

check/checker/types.go
```diff
@@ -19,3 +19,8 @@ import v1 "github.com/alibaba/sealer/types/api/v1"
 type Checker interface {
 	Check(cluster *v1.Cluster) error
 }
+
+// PreChecker: checking prerequisite for sealer,such as system info,register login status.
+type PreChecker interface {
```
fanux, line 24: I think we don't need this interface, just using Checker interface

fanux (approved):

PR #343, opened 2021-06-22
Optimize/not set aksk log

bxy4543:
Optimize not set aksk log

### DIFF
--- a/infra/infra.go
+++ b/infra/infra.go
@@ -15,8 +15,9 @@
 package infra
 
 import (
+	"fmt"
+
 	"github.com/alibaba/sealer/infra/aliyun"
-	"github.com/alibaba/sealer/logger"
 	v1 "github.com/alibaba/sealer/types/api/v1"
 )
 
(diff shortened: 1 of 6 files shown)


infra/infra.go
```diff
@@ -42,8 +43,8 @@ func NewDefaultProvider(cluster *v1.Cluster) Interface {
 		if err != nil {
 			logger.Error(err)
 		}
-		return aliProvider
+		return aliProvider, nil
 	default:
-		return nil
+		return nil, fmt.Errorf("the provider is invalid")
```
kakaZhou719, line 48: as my opinion, NewDefaultProvider is just a New method for struct ,shoud not return error, nil means New struct failed.

infra/infra.go
```diff
@@ -26,24 +25,27 @@ type Interface interface {
...
-		aliProvider.Config = *config
-		aliProvider.Cluster = cluster
-		err = aliProvider.NewClient()
-		if err != nil {
-			logger.Error(err)
-		}
-		return aliProvider
+		return NewAliProvider(cluster)
 	default:
-		return nil
```
fanux, line 47: I think default should return nil,error

infra/infra.go
fanux: Sometimes it will failed, like can't found AK

fanux (approved):

PR #323, opened 2021-06-16
WIP: Feature/buildcache

touchuyht:
### Describe what this PR does / why we need it
Build cache mechanism enables sealer to use the previous built image layers to increase image build speed.

### Does this pull request fix one issue?

Fixes #111 
### Describe how you did it
**imageID**: ImageID is calculated from the HASH256 hash of its discription yaml file.
**layerID**: Layers are referenced by cryptographic hashes of their serialized representation. This is a SHA256 digest over the tar archive used to transport the layer.

**StoreID**: StoreID is caculated from a series of serialized layers. The layers hash value is "", but the layers value and type is not empty. It is used in image store to help to reference to a stack of layers with a single identifier.

PR #346, opened 2021-06-23
Feature/buildcache

justadogistaken:
this is finishing pr for [#323](https://github.com/alibaba/sealer/pull/323)

### New Feature
This feature aims to bring layer cache ability to sealer build. The cache ability mainly disposes problems like waste of disk space, and repeating image id. By the way, this feature will bring more time consumption on sealer build, due to the cache id calculation.

There are some new structures and files that plays role in this feature.
cache Layer
```
type Layer struct {
	CacheID string `json:"cache_id"`
	Type    string `json:"type"`
	Value   string `json:"value"`
}
```
this "Layer" looks familiar to v1Layer, only CacheID and ID are different. ID of v1Layer is digest(tar(layer-files)). CacheID of cacheLayer is */layerdb/layerID/cacheID, which is calculated from src in `COPY src tar`.  CacheID is used to validate the consistency between src files from different `COPY src tar`, to see if it possible to hit the cache.

```
ChainID
```
I leant the The concept of ChainID from docker, but it's not complex like it's in docker. The key concept for ChainID is an entire file system over multiples layers. Which is computed over cache Layers.
The algorithm is:
```
func (l *Layer) ChainID(parentID ChainID) (ChainID, error) {
	if parentID.String() == "" {
		return ChainID(digest.FromString(l.String())), nil
	}
	return ChainID(digest.FromString(parentID.String() + ":" + l.String())), nil
}
```

We will use ChainID to hit cache, same ChainID means same file system.

### DIFF
--- a/image/store/filesystem.go
+++ b/image/store/filesystem.go
@@ -22,113 +22,132 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
-
-	"github.com/docker/docker/pkg/ioutils"
+	"strings"
+	"sync"
 
 	"github.com/alibaba/sealer/logger"
-
+	"github.com/docker/docker/pkg/ioutils"
 	"github.com/vbatts/tar-split/tar/asm"
 	"github.com/vbatts/tar-split/tar/storage"
 
-	"github.com/alibaba/sealer/utils"
+	"github.com/alibaba/sealer/common"
+	pkgutils "github.com/alibaba/sealer/utils"
+	"github.com/pkg/errors"
 
 	"github.com/opencontainers/go-digest"
 )
 
-type LayerStorage struct {
-	LayerDataRoot string
-	LayerDBRoot   string
+const (
+	imageDBRoot = common.DefaultImageDBRootDir
+)
+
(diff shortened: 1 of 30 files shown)


image/store/filesystem.go
```diff
@@ -22,96 +22,237 @@ import (
...
+	Get(id digest.Digest) ([]byte, error)
+	Set(data []byte) (digest.Digest, error)
+	Delete(id digest.Digest) error
+	ListImages() ([][]byte, error)
+	SetMetadata(id digest.Digest, key string, data []byte) error
+	GetMetadata(id digest.Digest, key string) ([]byte, error)
+	DeleteMetadata(id digest.Digest, key string) error
+	LayerDBDir(digest digest.Digest) string
+	LayerDataDir(digest digest.Digest) string
+	assembleTar(id LayerID, writer io.Writer) error
```
fanux, line 57: interface Using private function?

fanux (approved):

justadogistaken:
@touchuyht PTAL

Issue #353, opened 2021-06-25
Context useless for local build


justadogistaken:
I have a Kubefile:
```
FROM scratch
COPY kubeadm .
```
and I have kubeadm under ./bin

and I go sealer build -t test:v1 -c ./bin/

the result is error of "failed to exec layer &{ COPY kubeadm .}:stat kubeadm: no such file or directory"

fanux:
@kakaZhou719 You need to set the context dir as build work dir.

PR #355, opened 2021-06-28
fix:add build context

kakaZhou719:
1, add context path at local build
2, delete -c flag on build cmd

### Describe how to verify it
test report
https://github.com/alibaba/sealer/runs/2938256523?check_suite_focus=true
### Special notes for reviews
link to issue #353

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -101,7 +101,7 @@ const (
 	CdAndExecCmd    = "cd %s && %s"
 	TagImageCmd     = "%s tag %s %s"
 	PushImageCmd    = "%s push %s"
-	BuildClusterCmd = "%s build -f %s -t %s -c %s -b %s"
+	BuildClusterCmd = "%s build -f %s -t %s -b %s %s"
 )
 const ExecBinaryFileName = "sealer"
 const ROOT = "root"
--- a/common/common.go
+++ b/common/common.go
@@ -101,7 +101,7 @@ const (
 	CdAndExecCmd    = "cd %s && %s"
 	TagImageCmd     = "%s tag %s %s"
 	PushImageCmd    = "%s push %s"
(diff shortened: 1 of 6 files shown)


common/common.go
```diff
@@ -101,7 +101,7 @@ const (
 	CdAndExecCmd    = "cd %s && %s"
 	TagImageCmd     = "%s tag %s %s"
 	PushImageCmd    = "%s push %s"
-	BuildClusterCmd = "%s build -f %s -t %s -b %s ."
+	BuildClusterCmd = "%s build -f %s -t %s -c %s -b %s"
```
fanux, line 104: I think we don't need add -c? fllow Docker build command line?

kakaZhou719: if user set context dir ,we need to set it . when build context is not set, we can use  "."  by default.

fanux: Why not `sealer build -t xxx .` but `sealer build -t xxx -c .`?  We need context but needn't -c flag I think

sealer/cmd/build.go
```diff
@@ -25,18 +25,21 @@ import (
...
 
 // buildCmd represents the build command
 var buildCmd = &cobra.Command{
-	Use:     "build",
+	Use:     "build [flags] PATH",
 	Short:   "cloud image local build command line",
 	Example: `sealer build -f Kubefile -t my-kubernetes:1.18.3 .`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			logger.Error("sealer build requires exactly 1 argument.")
```
bxy4543, line 40: if args ==0 ，context = "."  ，  else context = arg0
                   i think

fanux (approved):

Issue #327, opened 2021-06-17
Specify the deployment address of the registry


ly-wjj:
## Issue Description

Type: *feature request*

### Describe what feature you want
The designated server can access the public network, we need to deploy the registry to the corresponding machine

lllwan:
Features have been merged: #377

PR #360, opened 2021-06-29
add feature Specify the registry server

ly-wjj:
The designated server can access the public network, we need to deploy the registry to the corresponding machine

 #327


fanux:
https://github.com/alibaba/sealer/pull/359 I think we can using a plugin to start a registry. Clusterfile upgrade should be deliberateness. Interface changes and standard changes need more discussion.

Like this:
```
kind: Plugin
metadata:
  name: SHELL
spec:
  action: PreInit
  on: node=192.168.0.2
  data: |
     docker start --name=registry --net=host -d registry:2.7
```

types/api/v1/cluster_types.go
```diff
@@ -59,6 +59,7 @@ type ClusterSpec struct {
 	CertSANS []string `json:"certSANS,omitempty"`
 	Masters  Hosts    `json:"masters,omitempty"`
 	Nodes    Hosts    `json:"nodes,omitempty"`
+	Registry string   `json:"registry,omitempty"`
```
fanux, line 62: Clusterfile adding fields requires review, and making as few changes as possible.

runtime/runtime.go
```diff
@@ -29,6 +29,7 @@ import (
 type Interface interface {
 	// exec kubeadm init
 	Init(cluster *v1.Cluster) error
+	InitRegistry(cluster *v1.Cluster) error
```
fanux, line 32: I think using a plugin to start the registry is better.

fanux:
https://github.com/alibaba/sealer/issues/374

fanux:
https://github.com/alibaba/sealer/pull/377  Using registry config file

PR #377, opened 2021-07-01
Support custom registry configuration

lllwan:


runtime/registry.go
```diff
@@ -16,21 +16,56 @@ package runtime
...
+}
+
+func (d *Default) getRegistryConfig() *RegistryConfig {
+	var config RegistryConfig
+	registryConfigPath := filepath.Join(common.DefaultClusterBaseDir(d.ClusterName), "etc/registry.yaml")
+	if utils.IsFileExist(registryConfigPath) {
+		err := utils.UnmarshalYamlFile(registryConfigPath, &config)
+		if err == nil && config.IP != "" && config.Domain != "" {
+			if config.Volume == "" {
+				config.Volume = "/var/lib/registry"
```
fanux, line 49: Volume=filepath.Join(DefaultClusterBaseDir(name),"/registry")

apply/cloud_applier.go
```diff
@@ -43,6 +45,7 @@ func NewAliCloudProvider(cluster *v1.Cluster) Interface {
 		ImageManager:   image.NewImageService(),
 		FileSystem:     filesystem.NewFilesystem(),
 		Guest:          guest.NewGuestManager(),
+		Config:         config.NewConfiguration(cluster.Name),
```
fanux, line 48: cloud mod using config is a litter complexity，need copy the Clusterfile configs to the remote server, I will do it later, delete this now.

runtime/registry.go
```diff
@@ -16,21 +16,56 @@ package runtime
...
+	}
+	return fmt.Sprintf("%s %s", cf.IP, cf.Domain)
+}
+
+func (d *Default) getRegistryConfig() *RegistryConfig {
+	var config RegistryConfig
+	registryConfigPath := filepath.Join(common.DefaultClusterBaseDir(d.ClusterName), "etc/registry.yaml")
+	if utils.IsFileExist(registryConfigPath) {
+		err := utils.UnmarshalYamlFile(registryConfigPath, &config)
+		if err == nil && config.IP != "" && config.Domain != "" {
```
fanux, line 47: I think we can set default domain and default host IP here.

apply/cloud_applier.go
lllwan: OK

runtime/registry.go
lllwan: default host IP defined in the getregistryhost() function. so, I don't think we need to change it.

runtime/registry.go
lllwan: If you need to do this, then the custom volume is meaningless. First remove the function of the custom volume.

runtime/registry.go
```diff
@@ -16,21 +16,58 @@ package runtime
...
+	if cf == nil {
+		return fmt.Sprintf("%s %s", d.Masters[0], SeaHub)
+	}
+	return fmt.Sprintf("%s %s", cf.IP, cf.Domain)
+}
+
+func (d *Default) getRegistryConfig() *RegistryConfig {
+	var config RegistryConfig
+	registryConfigPath := filepath.Join(common.DefaultClusterBaseDir(d.ClusterName), "etc/registry.yaml")
+	if utils.IsFileExist(registryConfigPath) {
```
fanux, line 44: if !utils.IsFileExist(registryConfigPath) {
   return nil
}

"return fast less retract"

lllwan: OK

fanux (approved):

fanux:
LGTM needs test

filesystem/filesystem.go
```diff
@@ -147,6 +149,11 @@ func (c *FileSystem) UnMountRootfs(cluster *v1.Cluster) error {
 
 func mountRootfs(ipList []string, target string, cluster *v1.Cluster) error {
 	SSH := ssh.NewSSHByCluster(cluster)
+	d := runtime.Default{
```
fanux, line 152: Using the runtime struct seems not very good, other package only care about the runtime interface.

I think we can define the function GetRegistryConfig(defaultIP string), so you don't need the runtime.Default struct here.

lllwan: OK

config/config.go
```diff
@@ -50,6 +50,7 @@ func NewConfiguration(clusterName string) Interface {
 
 func (c *Dumper) Dump(clusterfile string) error {
 	if clusterfile == "" {
+		logger.Warn("clusterfile is empty!")
```
fanux, line 53: If using run command clusterfile is "", but it normal, So I think using debug log here is better.

lllwan: OK

fanux (approved):

PR #400, opened 2021-07-06
[wip] feat: use grpc to proxy ssh cmd

justadogistaken:
未命名文件 (2)

There are many remote procedure call in our code. `ssh.CmdAsync(ip, cmd)` is how we do rpc, which lacks readability and failed to execute complicated logic on cmd feedback. Such way of rpc should be as few as possible in our project.
Therefore, I propose we use grpc to take place of `ssh.Cmd`. With the architecture I'm developing, we can do `mkdir -p /var/lib/sealer` in the following way:
```
clt, err := client.RemoteHandler(host, d.SSH)
if err != nil {
  // err handle
}

ctx := context.Background()
err = clt.OSCallSvc().Mkdir(ctx, file.DestinationDir, 0755)
if err != nil {
  // err handle
}
```

But I stuck in an issue of when to start sealer grpc server at remote machine. Currently I will check grpc liveness of host in client.RemoteHandler(host, d.SSH). sealer will send sealer or try to start grpc server if remote host didn't start the grpc server.


fanux:
So we need to run an engine on every node?

justadogistaken:
@fanux I suppose It's impossible to avoid.

justadogistaken:
@fanux @kakaZhou719 @cuisongliu PTAL

Issue #401, opened 2021-07-07
calico out of box


fanux:
Sealer now installs calico using a template, but the template is in [source code](https://github.com/alibaba/sealer/blob/main/guest/calico.go#L44) which is really not good.

Now sealer has the config feature, so we can completely not care about the CNI in source code, we treat it like a regular app.

When define a config in Clusterfile like this:

```
apiVersion: sealer.aliyun.com/v1alpha1
kind: Cluster
metadata:
  name: my-cluster
spec:
  image: registry.cn-qingdao.aliyuncs.com/sealer-app/my-SAAS-all-inone:latest
  provider: BAREMETAL
...
---
apiVersion: sealer.aliyun.com/v1alpha1
kind: Config
metadata:
  name: calico
spec:
  path: etc/calico-values.yaml
  data: |
      podcidr: x.x.x.x/xx
      IPIP: off
```
The config module will write the data to $rootfs/etc/calico-values.yaml (already done) 

So you just need to create a new Kubefile and build a calico CloudImage example:
```
FROM kubernetes:v1.19.9
COPY calico-crd .
CMD kubectl apply -f calico-crd
CMD kubectl apply -f etc/calico-values.yaml
```
`sealer build -t kubernetes-withcalico:v1.19.9 .`

Whats you needs to do?

1. Delete all the calico-related source code, like [v1.Cluster](https://github.com/alibaba/sealer/blob/main/types/api/v1/cluster_types.go#L28)  and apply CNI and calico template etc...
2. Write the calico Kubefile in https://github.com/alibaba/sealer/tree/main/applications/calico/vx.x.x
3. Build and test the CloudImage
4. Push the CloudImage to ACR hub

wuxingzhong:
I do.

PR #402, opened 2021-07-08
calico out of box

wuxingzhong:
###  What this PR does 

Remove CNI  from source code.  

###  Why we need it

Sealer now installs calico using a template, but the template is [source code](https://github.com/alibaba/sealer/blob/main/guest/calico.go#L44) which is really not good.

Now sealer has the config feature, so we can completely not care about the CNI in source code, we treat it like a regular app.

So remove CNI from the source code is the first step.

I will put CNI  plugin in applications in the next step. 
 
For more information, see:  [calico out of box](https://github.com/alibaba/sealer/issues/401)


justadogistaken:
@wuxingzhong Why would we remove the cni from sealer? Could you describe more details about it ?

wuxingzhong:
> @wuxingzhong Why would we remove the cni from sealer? Could you describe more details about it ?

ok, I have changed the description.

fanux:
@justadogistaken because #401

fanux (approved):

jafardu:
add cilium

jafardu:
> @justadogistaken because #401

add cilium

PR #421, opened 2021-07-14
get images from charts

wuxingzhong:


fanux:
The filename is wrong, charts not chars

build/lite/charts/chars.go
```diff
@@ -0,0 +1,61 @@
...
+		if len(images) != 0 {
+			list = append(list, images...)
+		}
+	}
+
+	list = removeDuplicate(list)
+	return list, nil
+}
+
+func NewChars() (Interface, error) {
```
fanux, line 42: NewCharts

fanux (approved):

PR #426, opened 2021-07-15
Lite build support

lllwan:


build/lite/charts/charts.go
```diff
@@ -34,8 +48,6 @@ func (charts *Charts) ListImages(clusterName string) ([]string, error) {
 			list = append(list, images...)
 		}
 	}
-
-	list = removeDuplicate(list)
```
fanux, line 38: Why delete this?

lllwan: Image pull will be de duplicated. Here, de duplication is equivalent to two operations. The actual pull is to parse the merge data of the helm chart and read the image. The de duplication here only deals with the data of helm chart

sealer/cmd/build.go
```diff
@@ -65,6 +68,7 @@ func init() {
 	buildCmd.Flags().StringVarP(&buildConfig.BuildType, "buildType", "b", "", "cluster image build type,default is cloud")
 	buildCmd.Flags().StringVarP(&buildConfig.KubefileName, "kubefile", "f", "Kubefile", "kubefile filepath")
 	buildCmd.Flags().StringVarP(&buildConfig.ImageName, "imageName", "t", "", "cluster image name")
+	buildCmd.Flags().StringVarP(&buildConfig.ImageName, "imageList", "i", "", "build by submitting image list")
```
fanux, line 71: Just using $rootfs/manifests/imageList file, for most users don't care about this argument.

fanux:
How do you save the new container images to the CloudImage layer?

sealer/cmd/build.go
lllwan: OK

fanux (approved):

PR #432, opened 2021-07-16
fix(main): add rootfs struct

cuisongliu:

### DIFF
--- a/filesystem/rootfs/docker/etc/daemon.json
+++ b/filesystem/rootfs/docker/etc/daemon.json
@@ -1,0 +1,20 @@
+{
+  "max-concurrent-downloads": 20,
+  "log-driver": "json-file",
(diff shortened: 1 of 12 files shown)


fanux (approved):

filesystem/rootfs/docker/etc/daemon.json
```diff
@@ -0,0 +1,20 @@
+{
+  "max-concurrent-downloads": 20,
+  "log-driver": "json-file",
+  "log-level": "warn",
+  "log-opts": {
+    "max-size": "10m",
```
oldthreefeng, line 6: max-size: 10m  seems to small for prodcution use. need to be set 100m or bigger.

oldthreefeng (commented):
init-kube.sh need to use bash to execute.

PR #435, opened 2021-07-16
optimize: add image store

justadogistaken:
This PR is for unify the interface for image/layer/metadata to operate get/delete/save.
Try not to GetImage simply from a global func like: `utils.GetImageByID`. Get image by image store instead.
```
func GetImageByID(imageID string) (*v1.Image, error) {
	fileName := filepath.Join(common.DefaultImageDBRootDir, imageID+".yaml")

	var image v1.Image
	return &image, utils.UnmarshalYamlFile(fileName, &image)
}
```
未命名文件 (1)


justadogistaken:
@touchuyht @fanux @kakaZhou719 @bxy4543 PTAL

bxy4543 (approved):

PR #446, opened 2021-07-23
Fix the returned error details,and add the output information of the …

liutao841204:
Fix the returned error details,This error is located in https://github.com/alibaba/sealer/blob/main/build/context.go#L54

### DIFF
--- a/build/context.go
+++ b/build/context.go
@@ -50,10 +50,10 @@ func (c *CloudBuilder) sendBuildContext() (err error) {
 	}()
 	// send to remote server
 	workdir := fmt.Sprintf(common.DefaultWorkDir, c.local.Cluster.Name)
-	err = c.SSH.Copy(c.RemoteHostIP, tarFileName, tarFileName)
-	if err != nil {
-		return err
+	if err = c.SSH.Copy(c.RemoteHostIP, tarFileName, tarFileName); err != nil {
+		return fmt.Errorf("failed to copy tar file: %s, err: %v", tarFileName, err)
 	}
+
 	// unzip remote context
 	err = c.SSH.CmdAsync(c.RemoteHostIP, fmt.Sprintf(common.UnzipCmd, workdir, tarFileName, workdir))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


build/context.go
```diff
@@ -50,9 +50,11 @@ func (c *CloudBuilder) sendBuildContext() (err error) {
 	}()
 	// send to remote server
 	workdir := fmt.Sprintf(common.DefaultWorkDir, c.local.Cluster.Name)
-	if err = c.SSH.Copy(c.RemoteHostIP, tarFileName, tarFileName); err != nil {
+	err = c.SSH.Copy(c.RemoteHostIP, tarFileName, tarFileName)
+	if err != nil {
 		return err
```
fanux, line 55: We want return error details, for example:

return fmt.Errorf("failed to copy tar file: %s, err: %v", tarFileName, err)

This format is fine
```
 if err =xxx; err != nil {
...
}
```

build/context.go
```diff
@@ -50,9 +50,11 @@ func (c *CloudBuilder) sendBuildContext() (err error) {
...
 	}
+	logger.Info("the program is sending the %s file!", tarFileName)
```
fanux, line 57: We needn't add log here, because the Copy funcion already has log.

liutao841204: OK

build/context.go
```diff
@@ -50,7 +50,8 @@ func (c *CloudBuilder) sendBuildContext() (err error) {
 	}()
 	// send to remote server
 	workdir := fmt.Sprintf(common.DefaultWorkDir, c.local.Cluster.Name)
-	if err = c.SSH.Copy(c.RemoteHostIP, tarFileName, tarFileName); err != nil {
+	err = c.SSH.Copy(c.RemoteHostIP, tarFileName, tarFileName)
```
fanux, line 53:
```suggestion
	if err = c.SSH.Copy(c.RemoteHostIP, tarFileName, tarFileName); err != nil {
	    return fmt.Errorf("failed to copy tar file: %s, err: %v", tarFileName, err)
	}
```

PR #462, opened 2021-07-26
[optimize] cache docker image 

kakaZhou719:
1,cache docker images while run build.
2,optimize build code:check build args such as: context and kubefile .make them as abs path.
3,optimize build code: split common code to utils.go
4, optimize build code: split build mount code to build_mount.go, support to add default target and upper.

### Describe how you did it
save docker images as a cluster image layer , which type is BASE and value is sha hash.

when start docker registry ,will merge and mount this base layer and rootfs regsitry first. when build k8s app , the Upper will cache docker images and save it with cluster image.


build/utils.go
```diff
@@ -0,0 +1,319 @@
...
+	if err != nil {
+		return ""
+	}
+	containers, err := cli.ContainerList(ctx, types.ContainerListOptions{})
+
+	if err != nil {
+		return ""
+	}
+
+	for _, c := range containers {
```
fanux, line 120: To many indentation

fanux: I think you can just using a default mount dir

image/util.go
```diff
@@ -42,6 +42,16 @@ func GetImageLayerDirs(image *v1.Image) (res []string, err error) {
 	return
 }
 
+//get docker image layer hash path
+func GetBaseDockerImageLayerDir(image *v1.Image) (res []string) {
```
justadogistaken, line 46: I guess base is the "diff" at build stage? especially registry cache layer. So I think "base" will make people feel confused. Docker has Base concept, but this is not "docker base", maybe "diff" "fsdiff" "build_diff" or something would be better.

image/util.go
```diff
@@ -42,6 +42,16 @@ func GetImageLayerDirs(image *v1.Image) (res []string, err error) {
...
+	for _, layer := range image.Spec.Layers {
+		if layer.ID != "" && layer.Type == common.BaseImageLayerType {
+			res = append(res, filepath.Join(common.DefaultLayerDir, layer.ID.Hex()))
```
justadogistaken, line 49: I recommend use LayerDataDir() function under Backend to get the path, do not only const directly.

image/util.go
```diff
@@ -42,6 +42,16 @@ func GetImageLayerDirs(image *v1.Image) (res []string, err error) {
 	return
 }
 
+//get docker image layer hash path
```
justadogistaken, line 45: this func is called outside pkg image. try use interface to provide service for outside pkg, or make it a local func for other pkg.

PR #463, opened 2021-07-27
optimize: reconstruct cmd/run/copy instruction condition control at build stage

justadogistaken:


justadogistaken:
There maybe a potential problem here.
We create hard links for /var/lib/sealer/data/{layer}... from build context, but if files under /var/lib/sealer/data/{layer}... change, it will propagate to build context. But in our usage scenes, the potential problem could be ignored, could it?
@fanux @kakaZhou719 help review the potential problem.

kakaZhou719:
@justadogistaken , yes,i think so . build context will be refreshed(delete if exist) every time , means the old changed layer will be ignored 。

justadogistaken:
@kakaZhou719 But if we modify the original files, the underlying layer data changed, which will result in many problems. 
I improved some build logic, ptal

fanux (approved):

fanux:
@bxy4543 need test

Issue #464, opened 2021-07-27
feature: generate log file.


justadogistaken:
Currently, with sealer cmd going, the executions will be output to terminal. But I didn't find some log files to keep those log info. I guess we may need file saved under /var/lib/sealer/log or something to record the executions.

fanux:
Maybe we can develop this feature in the logger package.

PR #465, opened 2021-07-28
add default log config

bxy4543:
The default log file is /root/.sealer.json or something else?

#464

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -24,6 +24,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/spf13/viper"
+
 	"github.com/alibaba/sealer/common"
 )
 
@@ -132,7 +134,7 @@ func NewLogger(depth ...int) *LocalLogger {
 	l.appName = "[" + appSn + "]"
 	l.callDepth = dep
 	l.usePath = true
-	l.SetLogger(AdapterConsole)
+	//l.SetLogger(AdapterConsole)
 	l.timeFormat = logTimeDefaultFormat
 	return l
 }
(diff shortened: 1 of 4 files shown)


logger/log.go
```diff
@@ -149,6 +153,23 @@ func init() {
 	defaultLogger = NewLogger(3)
 }
 
+func Cfg(configFile string, debugMod bool, logFile string) {
+	var logCfg logConfig
+	viper.SetConfigType("json")
+	err := viper.Unmarshal(&logCfg)
+	if err != nil {
+		cfg, _ := json.Marshal(&logCfg)
```
justadogistaken, line 161: ~/.sealer/config.json is just a string? not a concrete struct? what does the config look like?

logger/log.go
```diff
@@ -362,22 +383,57 @@ func SetLogger(param ...string) {
...
-		fd, err := os.Open(c)
-		if err != nil {
-			fmt.Fprintf(common.StdErr, "Could not open %s for configure: %s\n", c, err)
-			os.Exit(1)
-		}
+		//configuration file does not exist, use the default configuration
+		if f, err := os.Stat(c); err != nil || f.IsDir() {
+			var (
+				loglevel   = 5
+				loggerfile = filepath.Join(common.GetHomeDir(), ".sealer", "sealer.log")
```
justadogistaken, line 390: and how about we save log with more info? sealer-2011-1-1-13-13-13.log or something?

logger/log.go
bxy4543: {"TimeFormat":"2006-01-02 15:04:05","Console":{"level":"","color":true,"LogLevel":5},"File":{"filename":"/root/.sealer/sealer.log","append":true,"maxlines":1000000,"maxsize":1,"daily":false,"maxdays":-1,"level":"TRAC","permit":"0660","LogLevel":0}}
https://github.com/alibaba/sealer/blob/f7218f721e9ca413e1db9a05722a55a72952774b/logger/log_test.go#L56-L70

sealer/cmd/root.go
```diff
@@ -51,28 +51,21 @@ func Execute() {
 
 func init() {
 	cobra.OnInitialize(initConfig)
-	rootCmd.PersistentFlags().StringVar(&rootOpt.cfgFile, "config", "", "config file (default is $HOME/.sealer.yaml)")
+	rootCmd.PersistentFlags().StringVar(&rootOpt.cfgFile, "config", "", "config file (default is $HOME/.sealer.json)")
```
fanux, line 54: I think we don't need to add a config file, just save logs in ${rootfs}/var/log/xxx.log.

bxy4543: If the configuration file does not exist, the default configuration is used. The default log file path is similar to /var/log/sealer/2006-01-01.log

fanux (approved):

PR #475, opened 2021-08-03
add the build parameter to support submitting the private registry ac…

lllwan:
…count password

### DIFF
--- a/build/lite/docker/pull.go
+++ b/build/lite/docker/pull.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/alibaba/sealer/image/reference"
+
 	"github.com/alibaba/sealer/common"
 	dockerstreams "github.com/docker/cli/cli/streams"
 	dockerjsonmessage "github.com/docker/docker/pkg/jsonmessage"
@@ -32,6 +34,7 @@ import (
 )
 
 type Docker struct {
+	Auth     string
 	Username string
 	Password string
 }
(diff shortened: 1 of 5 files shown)


justadogistaken:
@lllwan And we can read from ~/.docker/config.json to see if the auth info already exists.

build/lite/docker/pull.go
```diff
@@ -60,27 +63,38 @@ func (d Docker) ImagesPullByList(images []string) {
...
+		logger.Warn("docker client creation failed: %v", err)
 		return err
 	}
-	if d.Username != "" && d.Password != "" {
-		authConfig := types.AuthConfig{
-			Username: d.Username,
-			Password: d.Password,
-		}
+	authConfig, err = utils.GetDockerAuthInfoFromDocker(named.Domain())
+	if err != nil {
```
justadogistaken, line 86: err != nil ?

justadogistaken: I think we can tolerate the fault here, because we could mount config into registry.

lllwan: OK

build/lite/docker/pull.go
```diff
@@ -60,27 +63,37 @@ func (d Docker) ImagesPullByList(images []string) {
...
-			Username: d.Username,
-			Password: d.Password,
-		}
-		encodedJSON, err := json.Marshal(authConfig)
-		if err != nil {
-			return err
-		}
-		authStr := base64.URLEncoding.EncodeToString(encodedJSON)
-		ImagePullOptions = types.ImagePullOptions{RegistryAuth: authStr}
+	authConfig, _ = utils.GetDockerAuthInfoFromDocker(named.Domain())
```
justadogistaken, line 86: we still need the error, if error happens, just pass a empty RegistryAuth.

lllwan: The GetDockerAuthInfoFromDocker function handles the error and must return an auth object

justadogistaken (approved):

PR #480, opened 2021-08-04
optimize build code

kakaZhou719:

1,optimize build code:check build args such as: context and kubefile .make them as abs path.
2,optimize build code: split common code to utils.go
3, optimize build code: split build mount code to build_mount.go, support to add default target and upper.

### DIFF
--- a/build/build_mount.go
+++ b/build/build_mount.go
@@ -1,0 +1,99 @@
+package build
+
+import (
+	"fmt"
+
+	"github.com/alibaba/sealer/logger"
+	"github.com/alibaba/sealer/utils"
+	"github.com/alibaba/sealer/utils/mount"
+)
+
+type MountTarget struct {
+	driver     mount.Interface
+	TempTarget string
+	TempUpper  string
+	LowLayers  []string
+}
+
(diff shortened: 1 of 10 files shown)


build/build_mount.go
```diff
@@ -0,0 +1,108 @@
...
+
+type MountTarget struct {
+	driver     mount.Interface
+	TempTarget string
+	TempUpper  string
+	LowLayers  []string
+}
+
+func (m MountTarget) TempMount() error {
+	fmt.Println(m.TempTarget, m.TempUpper, m.LowLayers)
```
fanux, line 28: Using logger instead

build/build_mount.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"fmt"
+
+	"github.com/alibaba/sealer/logger"
+	"github.com/alibaba/sealer/utils"
+	"github.com/alibaba/sealer/utils/mount"
+)
+
+type IMountDriver interface {
```
fanux, line 11: What is this interface for? I think you need add some comment

fanux (approved):

PR #504, opened 2021-08-13
add apply precheck

### LINKED ISSUE #503: 安装过程中俩个报错
1.error execution phase upload-config/kubelet: Error writing Crisocket information for the control-plane node: timed out waiting for the condition
installer 192.168.0.10
installer 192.168.0.10
installer 192.168.0.10

bxy4543:
add apply precheck
add hostname check

fix #503   The last kubelet.conf was used because the environment was not cleaned up after the last execution.

https://github.com/alibaba/sealer/blob/6549495ad4ad4272cce80f341a0b5e0a2d245a4f/cert/kubeconfig.go#L277

### DIFF
--- a/apply/default_applier.go
+++ b/apply/default_applier.go
@@ -17,6 +17,9 @@ package apply
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
+
+	"github.com/alibaba/sealer/check/service"
 	"github.com/alibaba/sealer/common"
 	"github.com/alibaba/sealer/config"
 	"github.com/alibaba/sealer/filesystem"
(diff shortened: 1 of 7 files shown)


apply/default_applier.go
```diff
@@ -203,7 +208,7 @@ func (c *DefaultApplier) diff() (todoList []ActionName, err error) {
 		todoList = append(todoList, CleanFS)
 		return todoList, nil
 	}
-
+	todoList = append(todoList, ApplyCheck)
```
fanux, line 211: What if the checker is post check?

PR #510, opened 2021-08-18
bugfix:fix copy dir is "."

kakaZhou719:

### DIFF
--- a/build/instruction_handler.go
+++ b/build/instruction_handler.go
@@ -164,7 +164,7 @@ func (h *handler) handleCMDRUNCmd(layer v1.Layer, lowerLayers ...string) (layerI
 func (h *handler) copyFiles(srcFileName, dstFileName, tempBuildDir string) error {
 	var (
 		src = filepath.Join(h.hc.buildContext, srcFileName)
-		dst = tempBuildDir
+		dst string
 	)
 
 	fi, err := os.Stat(src)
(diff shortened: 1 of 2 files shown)


build/instruction_handler.go
```diff
@@ -173,10 +173,15 @@ func (h *handler) copyFiles(srcFileName, dstFileName, tempBuildDir string) error
 	}
 
 	if fi.IsDir() {
+		//default workdir is rootfs,so if copy dst is ".", name it as filepath.Base(srcFileName)
```
fanux, line 176: If dst dir is not absolute path， using filepath.join($rootfs, dst), else using dst dir directly

kakaZhou719: yes,we assume tempBuildDir is temp rootfs.so, the dst dir is filepath.join(tempBuildDir , dst).
if copy dir : 
1,copy . . :  src is context, dst is $rootfs
2,copy dir . : src is context/dir , dst is  $rootfs.
3,copy dirA dirB ： src is context/dirA , dst is  $rootfs/dirA/dirB

fanux: COPY dir .       dst is $rootfs/dir is better I think. = cp -r dir ./
COPY dirA ./dirB dst is $rootfs/dirB/dirA is better. = cp -r dirA ./dirB
COPY dirA /dirB dst is /dirB/dirA = cp -r dirA /dirB

kakaZhou719:
close it as we dont support system command now.

PR #511, opened 2021-08-18
fix plugin not working in cloud apply

bxy4543:
fix plugin not working in cloud apply
fix failed to umount $rootfs/registry cause the sealer delete failure

image

### DIFF
--- a/apply/cloud_applier.go
+++ b/apply/cloud_applier.go
@@ -15,20 +15,23 @@
 package apply
 
 import (
+	"bytes"
 	"fmt"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/alibaba/sealer/common"
+	"github.com/alibaba/sealer/config"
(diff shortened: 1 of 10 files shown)


apply/cloud_applier.go
```diff
@@ -132,9 +132,19 @@ func (c *CloudApplier) Apply() error {
 	}
 
 	cluster.Spec.Provider = common.BAREMETAL
-	err = utils.MarshalYamlToFile(common.TmpClusterfile, cluster)
-	if err != nil {
-		return fmt.Errorf("marshal tmp cluster file failed %v", err)
+	clusterFile := cluster.GetAnnotationsByKey(common.ClusterfileName)
```
fanux, line 135: I think we should generate the tmpClusterfile which append Config and Plugin.

1. generateTmpClusterfile()
       1. dump v1.Cluster
       2. append v1.Config
       3. append v1.Plugin
2. sendTmpClusterfile()

fanux (approved):

PR #528, opened 2021-08-24
Bugfix/infra container nil

hellomyboy:
find some error and want be developer

https://github.com/alibaba/sealer/issues/518


apply/default_applier.go
```diff
@@ -199,7 +199,10 @@ func (c *DefaultApplier) Apply() (err error) {
 		logger.Debug("sealer apply process %s", action)
 		err := ActionFuncMap[action](c)
 		if err != nil {
-			return err
+			if c.ClusterDesired.DeletionTimestamp == nil {
```
fanux, line 202: The registry dir umount error we will fix it in runtime.reset. I think we should not ignore all the delete errors

fanux: We already fixed this issue, I think you can just delete this

hellomyboy: ok i delete this

fanux (approved):

PR #534, opened 2021-08-25
E2e/join support consecutive ip

hellomyboy:
join sub command support consecutive IP  range, example : sealer join -m xxx.xxx.xxx.1-xxx.xxx.xxx.5 -n xxx.xxx.xxx.11-xxx.xxx.xxx.15


fanux:
delete the .DS_Store file

apply/run.go
```diff
@@ -61,6 +63,47 @@ func IsIPList(args string) bool {
...
+	}
+	return big.NewInt(0).SetBytes(ip.To16())
+}
+
+func CompareIP(v1, v2 string) int {
+	i := ipToInt(v1)
+	j := ipToInt(v2)
+
+	if i == nil || j == nil {
+		os.Exit(-1)
```
fanux, line 84: should not Exit, return error instead.

apply/run.go
```diff
@@ -61,6 +63,47 @@ func IsIPList(args string) bool {
 	return true
 }
 
+func PreProcessIPList(joinArgs *common.RunArgs) {
```
fanux, line 66: Provide some test cases for this function.
Attention about the edge cases like: `x.x.x.x-` x.x.x.x-x.x.x.x-x.x.x.x` ...

Issue #551, opened 2021-08-30
shell plugin support on field.


fanux:
```
apiVersion: sealer.aliyun.com/v1alpha1
kind: Plugin
metadata:
  name: SHELL
spec:
  action: PostInstall
  on: 192.168.0.2
  data: |
     kubectl taint nodes node-role.kubernetes.io/master=:NoSchedule
```

If `spec.on` is not null, exec the shell command on that node, else exec it on every node.

https://github.com/alibaba/sealer/blob/main/plugin/shell_plugin.go

hellomyboy:
ok,i will support it 

> ```
> apiVersion: sealer.aliyun.com/v1alpha1
> kind: Plugin
> metadata:
>   name: SHELL
> spec:
>   action: PostInstall
>   on: 192.168.0.2
>   data: |
>      kubectl taint nodes node-role.kubernetes.io/master=:NoSchedule
> ```
> 
> If `spec.on` is not null, exec the shell command on that node, else exec it on every node.
> 
> https://github.com/alibaba/sealer/blob/main/plugin/shell_plugin.go

PR #559, opened 2021-09-02
shell plugin support on field. #551

x-shadow-man:
shell plugin support on field. #551

### DIFF
--- a/apply/default_applier.go
+++ b/apply/default_applier.go
@@ -259,8 +259,6 @@ func (c *DefaultApplier) diff() (todoList []ActionName, err error) {
 	}
 	todoList = append(todoList, Guest)
 	todoList = append(todoList, UnMountImage)
-	todoList = append(todoList, PluginDump)
-	todoList = append(todoList, PluginPhasePostInstallRun)
 	return todoList, nil
 }
 
--- a/apply/default_applier.go
+++ b/apply/default_applier.go
@@ -259,6 +259,8 @@ func (c *DefaultApplier) diff() (todoList []ActionName, err error) {
 	}
 	todoList = append(todoList, Guest)
 	todoList = append(todoList, UnMountImage)
+	todoList = append(todoList, PluginDump)
+	todoList = append(todoList, PluginPhasePostInstallRun)
 	return todoList, nil
 }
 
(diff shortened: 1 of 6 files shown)


apply/default_applier.go
```diff
@@ -259,6 +259,8 @@ func (c *DefaultApplier) diff() (todoList []ActionName, err error) {
 	}
 	todoList = append(todoList, Guest)
 	todoList = append(todoList, UnMountImage)
+	todoList = append(todoList, PluginDump)
+	todoList = append(todoList, PluginPhasePostInstallRun)
```
fanux, line 263: plugin post install already in the currentcluster == nil block.

plugin/shell_plugin.go
```diff
@@ -35,10 +37,16 @@ func (s Sheller) Run(context Context, phase Phase) error {
 	//get all host ip
 	masterIP := context.Cluster.Spec.Masters.IPList
 	nodeIP := context.Cluster.Spec.Nodes.IPList
-	hostIP := append(masterIP, nodeIP...)
+	allHostIP := append(masterIP, nodeIP...)
+	//get on
+	on := context.Plugin.Spec.On
+	if on != "" {
+		IPList := utils.DisassembleIPList(on)
```
fanux, line 44: allHostIP = utils.DisassembleIPList(on)

plugin/shell_plugin.go
```diff
@@ -35,10 +37,16 @@ func (s Sheller) Run(context Context, phase Phase) error {
 	//get all host ip
 	masterIP := context.Cluster.Spec.Masters.IPList
 	nodeIP := context.Cluster.Spec.Nodes.IPList
-	hostIP := append(masterIP, nodeIP...)
+	allHostIP := append(masterIP, nodeIP...)
+	//get on
+	on := context.Plugin.Spec.On
+	if on != "" {
```
bxy4543, line 43: What about on : role=master or on : role=node

apply/default_applier.go
x-shadow-man: This func only works in installing phrase. if I want to apply this feature after installing k8s, it will not work.

plugin/shell_plugin.go
hellomyboy: have two pr，next pr will support it

plugin/shell_plugin.go
x-shadow-man: done. thx

apply/default_applier.go
fanux: Some plugin exec when the upgrade may occur some error,  like yum install some already exist tools.  But if not add this the new node will not exec the plugin command...

fanux (approved):

PR #569, opened 2021-09-06
Bugfix/reinstall mount failed

bxy4543:
fix sealer container failed to delete and join.


apply/delete.go
```diff
@@ -29,7 +29,9 @@ func Delete(cluster *v1.Cluster, scaleArgs *common.RunArgs) error {
 	case common.BAREMETAL:
 		return deleteBaremetalNodes(cluster, scaleArgs)
 	case common.AliCloud:
-		return deleteAliCloudNodes(cluster, scaleArgs)
+		return deleteCloudNodes(cluster, scaleArgs)
```
fanux, line 32: I think deleteInfraNodes is better, because container is not Cloud

fanux (approved):

PR #593, opened 2021-09-14
Ssh bug fix

lllwan:


fanux:
Maybe you can describe what the bug is?

fanux (approved):

lllwan:
> Maybe you can describe what the bug is?

We encountered a special scenario: it is not allowed to use the default port 22 for SSH communication. It is found that even if I specify the actual port, sealer will still try to use port 22, resulting in failure of communication.

bxy4543:
test apply

PR #617, opened 2021-09-16
move some modules to the pkg dir

ByboyCn:
move some modules to the pkg dir


apply/default_applier.go
```diff
@@ -36,9 +36,9 @@ import (
 type DefaultApplier struct {
 	ClusterDesired  *v1.Cluster
 	ClusterCurrent  *v1.Cluster
-	ImageManager    image.Service
+	ImageManager    image2.Service
```
fanux, line 39: Why rename the package name?

build/instruction_handler.go
```diff
@@ -38,9 +36,9 @@ import (
 type handlerContext struct {
 	buildContext  string
 	continueCache bool
-	cacheSvc      cache.Service
+	cacheSvc      cache2.Service
 	prober        image.Prober
-	parentID      cache.ChainID
+	parentID      cache2.ChainID
```
fanux, line 41: i think we don't need to rename the package name.

fanux:
I think you can PR those modules one by one?

Issue #600, opened 2021-09-15
Sealer login failed to private docker registry with https


jkryanchou:
#### What happened:

I have tried `sealer login`  to our company's private docker-registry which built on Harbor  as below 

```
sealer login registry.starx.com -uadmin -pHarbor12345
```

While it failed with the errors:

```
2021-09-15 13:23:48 [EROR] [default_image.go:165] registry.starx.com authentication failed
2021-09-15 13:23:48 [EROR] [login.go:51] Get "https://registry.starx.com/v2/": x509: certificate signed by unknown authority
```

#### What you expected to happen:

I did the same login-attempt through docker-client and it succeed without any errors

```
[root@localhost sealer-examples]# docker login registry.starx.com -uadmin -pHarbor12345
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
WARNING! Your password will be stored unencrypted in /root/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded
```

#### Environment:
- sealer version (use `sealer version`): {"gitVersion":"v0.4.0","gitCommit":"7341634","buildDate":"2021-08-31 10:45:26","goVersion":"go1.14.15","compiler":"gc","platform":"linux/amd64"}
- Cloud provider or hardware configuration: 
- OS (e.g: `cat /etc/os-release`):
  ```
   NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
  ```
- Kernel (e.g. `uname -a`): Linux localhost.localdomain 3.10.0-1127.el7.x86_64 #1 SMP Tue Mar 31 23:36:51 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
- Install tools: 
- Others:
   
   * My docker daemon configuration (`cat /etc/docker/daemon.json`) 
   
[...]

PR #618, opened 2021-09-17
Bugfix/lite fix

bxy4543:
Change the NewRegistryCache func order
fix sealer login x509 err : skip insecure verify.

#600

### DIFF
--- a/image/distributionutil/login.go
+++ b/image/distributionutil/login.go
@@ -16,8 +16,10 @@ package distributionutil
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
@@ -41,7 +43,9 @@ func Login(ctx context.Context, authConfig *types.AuthConfig) error {
 	}
 
 	modifiers := dockerRegistry.Headers(dockerversion.DockerUserAgent(ctx), nil)
-	authTransport := transport.NewTransport(dockerRegistry.NewTransport(nil), modifiers...)
+	base := dockerRegistry.NewTransport(nil)
+	base.TLSClientConfig.InsecureSkipVerify = os.Getenv("SKIP_TLS_VERIFY") == "true"
+	authTransport := transport.NewTransport(base, modifiers...)
 
 	credentialAuthConfig := *authConfig
 	creds := loginCredentialStore{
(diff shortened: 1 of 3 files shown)


image/distributionutil/login.go
```diff
@@ -41,7 +44,24 @@ func Login(ctx context.Context, authConfig *types.AuthConfig) error {
 	}
 
 	modifiers := dockerRegistry.Headers(dockerversion.DockerUserAgent(ctx), nil)
-	authTransport := transport.NewTransport(dockerRegistry.NewTransport(nil), modifiers...)
+	tlsConfig := tlsconfig.ServerDefault()
+	//skip insecure verify
+	tlsConfig.InsecureSkipVerify = true
```
justadogistaken, line 49: does the following config copy from dockerRegistry.NewTransport(nil)? if it does, we could get transport from dockerRegistry.NewTransport(nil), and change its TLSClientConfig.InsecureSkipVerify.
And is it a good solution to skip insecure verification always?

bxy4543: Can the environment variable 'SKIP_TLS_VERIFY=true' be used to determine whether to skip?

image/distributionutil/login.go
```diff
@@ -41,7 +46,21 @@ func Login(ctx context.Context, authConfig *types.AuthConfig) error {
...
+			CertFile: filepath.Join(dockerCertPath, "cert.pem"),
+			KeyFile:  filepath.Join(dockerCertPath, "key.pem"),
+		}
+		tlsc, err = tlsconfig.Client(options)
+		if err != nil {
+			return err
+		}
+	}
+	base := dockerRegistry.NewTransport(tlsc)
+	base.TLSClientConfig.InsecureSkipVerify = os.Getenv("SKIP_TLS_VERIFY") == "true"
```
justadogistaken, line 62: It's good, but do it before getting "DOCKER_CERT_PATH". Because if user decides to skip tls verify, why do we fetch cert?
Or skip tls and retrieve cert are not the same thing to do ?

justadogistaken:
@bxy4543  Skipping tls verification can make the client trust server cert, but I don't really get the point for providing docker cert, why should we handle this? Or export the env to user?

image/distributionutil/login.go
fanux: So user must set SKIP_TLS_VERIFY env to skip the verify issue? We can show some log if the error occured

bxy4543:
> @bxy4543 Skipping tls verification can make the client trust server cert, but I don't really get the point for providing docker cert, why should we handle this? Or export the env to user?

I was wrong, I thought it would affect the sealer pull image or something

justadogistaken:
@bxy4543  "strings.Contains(err.Error(), "x509: certificate signed by unknown authority")" 
It's ok. But is there any error type in net/http/...or other pkg?

bxy4543:
> @bxy4543 "strings.Contains(err.Error(), "x509: certificate signed by unknown authority")"
> It's ok. But is there any error type in net/http/...or other pkg?

I didn't find any type of x509 error in other pkg.

justadogistaken:
LGTM

justadogistaken (approved):

fanux (approved):

PR #623, opened 2021-09-17
Optimize save registry cache.

bxy4543:
Optimize save registry cache.
Change log file to /var/lib/sealer/log, because github action no permission to save logs when run e2e test.

### DIFF
--- a/utils/ssh/ssh.go
+++ b/utils/ssh/ssh.go
@@ -80,24 +80,26 @@ type Client struct {
 }
 
 func NewSSHClientWithCluster(cluster *v1.Cluster) (*Client, error) {
-	var host string
-	if cluster.Spec.Provider == common.CONTAINER {
-		host = cluster.Spec.Masters.IPList[0]
-	}
+	var (
+		ipList []string
+		host   string
+	)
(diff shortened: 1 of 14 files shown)


utils/ssh/ssh.go
```diff
@@ -80,40 +80,47 @@ type Client struct {
...
-	}
-
-	if host == "" {
-		return nil, fmt.Errorf("get cluster EIP failed")
+	err := WaitSSHReady(sshClient, 6, ipList...)
+	if err != nil {
+		return nil, err
 	}
-	err := WaitSSHReady(sshClient, host)
+	uLimit := `if [ "$(ulimit -n)" -lt "65535" ];then echo "* - nofile 65535" >> /etc/security/limits.conf; fi`
```
fanux, line 101: Add ulimit in NewSSHClient is very not friendly, The best way to config ulimit is using plugin.

bxy4543: > Add ulimit in NewSSHClient is very not friendly, The best way to config ulimit is using plugin.

well

kakaZhou719 (approved):

PR #632, opened 2021-09-18
Docs/arch

fanux:


allencloud:
This documents in the only source file to compose up the website?  I found most detailed content of the document is duplicated with the raw document which is not in the website folder. 

In my opinion, currently the website maintaining may cost some energy of the developers, even not some, but much. 

I see that the website increase the user experience a lot. While we should always emphasize the most important core of sealer's open source. With the core built, never concern about the spreading of value. But it should be taken care of when leveraging the sealer core team's resource to many areas. @fanux

fanux:
image

I think we should move `sealer/docs` to `docs/site/content/docs`, the site is a better way to show sealer docs.

For developers contribute the docs are quite simple, just add markdown into `docs/site/content/docs`, they needn't care about website render.

contribute guide: https://github.com/alibaba/sealer/issues/628

bxy4543 (approved):

PR #708, opened 2021-09-24
chore:remove docker client and k8s client to its dir

kakaZhou719:
as we see, sealer will use many client such as k8s client and docker client. so i move it to its own dir with the name. 

- 1, move k8s client 
- 2, move docker client

### DIFF
--- a/client/k8s/client.go
+++ b/client/k8s/client.go
@@ -28,7 +28,7 @@ import (
 	"k8s.io/client-go/util/homedir"
 )
 
-type K8sClient struct {
+type Client struct {
 	client *kubernetes.Clientset
 }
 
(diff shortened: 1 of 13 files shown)


client/k8s/client.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package client
+package k8s
```
bxy4543, line 15: K8s. Client is better than K8S. k8sClient
```
type K8sClient struct {
	client *kubernetes.Clientset
}
```

kakaZhou719: good reviews

bxy4543 (approved):

PR #726, opened 2021-09-28
update readme

bxy4543:
update readme


kakaZhou719:
cloud we do this automatically ? i read the source code, just modify the sealer version.

fanux (approved):

PR #727, opened 2021-09-29
feature:add imageList to local build

kakaZhou719:
1, add imageList to local build.
2, refactor lite build.


fanux:
I think imageList can separate from liteBuild or container build or cloudBuild. Using an independent build stage to handle imageList. So the container/lite/cloud build need not care about imageList.

kakaZhou719:
> I think imageList can separate from liteBuild or container build or cloudBuild. Using an independent build stage to handle imageList. So the container/lite/cloud build need not care about imageList.

yes, it is better to process the layer content in the build stage, 
meanwhile, to adapting to buildkit process, I modified lite build process.

PR #730, opened 2021-09-29
fix null point error caused by using label plugin.

bxy4543:
fix null point error caused by using label plugin.

### DIFF
--- a/utils/ssh/ssh.go
+++ b/utils/ssh/ssh.go
@@ -116,7 +116,8 @@ func WaitSSHReady(ssh Interface, tryTimes int, hosts ...string) error {
 		go func(host string) {
 			defer wg.Done()
 			for i := 0; i < tryTimes; i++ {
-				if err = ssh.Ping(host); err == nil {
+				err = ssh.Ping(host)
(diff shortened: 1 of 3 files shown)


common/common.go
```diff
@@ -89,6 +89,13 @@ const (
 	CRDCluster = "Cluster"
 )
 
+const (
+	EtcdPlugin     = "ETCD"
+	LabelPlugin    = "LABEL"
```
kakaZhou719, line 94: I think it's better to do this： move those  const to plugin dir

utils/ssh/ssh.go
```diff
@@ -116,7 +116,8 @@ func WaitSSHReady(ssh Interface, tryTimes int, hosts ...string) error {
 		go func(host string) {
 			defer wg.Done()
 			for i := 0; i < tryTimes; i++ {
-				if err = ssh.Ping(host); err == nil {
+				err = ssh.Ping(host)
```
kakaZhou719, line 119: is this change related to a bug?

fanux (approved):

PR #735, opened 2021-09-30
[sealer-app-images] add flannel images

bryant-rh:
[sealer-app-images] add flannel images

### DIFF
--- a/applications/flannel/init-kube.sh
+++ b/applications/flannel/init-kube.sh
@@ -1,0 +1,162 @@
+#!/bin/bash
+# Open ipvs
+modprobe -- ip_vs
+modprobe -- ip_vs_rr
+modprobe -- ip_vs_wrr
+modprobe -- ip_vs_sh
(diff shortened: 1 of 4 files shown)


applications/flannel/init-kube.sh
```diff
@@ -0,0 +1,162 @@
+#!/bin/bash
```
cuisongliu, line 1: why need this files ???? this file in the mini rootfs is already exsits

fanux:
Maybe you can provide a CloudImage address in readme, so others just pull and use it, not care about how to create a flannel CloudImage~

We also will sync the flannel CloudImage to the official registry.

bryant-rh:
> Maybe you can provide a CloudImage address in readme, so others just pull and use it, not care about how to create a flannel CloudImage~
> 
> We also will sync the flannel CloudImage to the official registry.

Ok, let me modify it！

Issue #775, opened 2021-10-19
inconsistent arguments of three subcommands


shy-Xu:
## Issue Description
sealer delete  is inconsistent with sealer join and sealer upgrade when there is an argument stands for cluster name.

e.g. sealer delete -m x.x.x.x -f /root/.sealer/my-cluster/Clusterfirl while sealer join -m x.x.x.x -c my-cluster 

I think it's better if we have: sealer delete -m x.x.x.x -c my-cluster.So that the three subcommands "delete"、"join"、"upgrade" will be consistent

PR #777, opened 2021-10-19
make delete command argument consistent

shy-Xu:
make delete cmd argument consistent

this PR is aimed for fixing the problem in the issue #775"inconsistent arguments of three subcommands"


shy-Xu:
I remain the -f  and add -c .

Issue #695, opened 2021-09-22
sealer ApplicationsImage & incremental updating


fanux:
# What is sealer ApplicationImage

Applications Image contains applications with all dependencies except the base image, so ApplicationsImage can install to an already existing Kubernetes cluster.

![image](https://user-images.githubusercontent.com/8912557/134311605-49efc418-7c7a-406a-a6fa-a23a8d304879.png)

`ApplicationImage = CloudImage - BaseImage`

# Why we need ApplicationsImage

Incremental updating & Install Applications to an already existing Kuberentes cluster.

# How to build an ApplicationImage

For example to build a Prometheus ApplicationImage:
Kubefile:
```
FROM registry.cn-qingdao.aliyuncs.com/sealer-apps/openebs-localpv:2.11.0
COPY prometheus .
CMD kubectl apply -f prometheus/crd.yaml
CMD kubectl apply -f prometheus/operator.yaml
```

```
sealer build -t prometheus:2.30.0 . --without-base
```
The prometheus:2.30.0 will not contain the embedded Kubernetes cluster.

# Difference from Docker image

The prometheus:2.30.0 ApplicationsImage contains all the DockerImage that Prometheus-stack needs, and contains helm chart or other operator manifests.

# How to Run ApplicationImage to an existing cluster

```
sealer run prometheus:2.30.0
```
That's all...

Or using Clusterfile to overwrite the config files like helm values.

# What if the cluster is not installed by sealer

Sealer hacked docker and registry to cache docker image in cluster, so if cluster not installed by sealer, how can we solve the Docker image cache issue.  We can only support sealer cluster now.

# Workflow

On build:
    Skip BaseImage and the `FROM` layer in metadata.
On Run:
   1. mount ApplicationImage Registry dir to sea.hub registry
   2. exec all CMD command

ningpengtao-coder:
以应用为粒度做增量更新，如果保证能够正确更新。例如执行CMD helm install xxx话，增量更新时需要执行helm upgrade，是要增加条件判断语句去更新吗？

kakaZhou719:
@ningpengtao-coder ，应该是后者的流程。就是只要我们能够保存升级的上下文和对应的容器镜像，应该就能保证在执行增量更新命令的时候成功升级。

ningpengtao-coder:
[...]

PR #817, opened 2021-11-02
feature: add without base option for sealer build

kakaZhou719:

Applications Image contains applications with all dependencies except the base image, so ApplicationsImage can install to an already existing Kubernetes cluster.

add "--no-base" option for sealer build, if it is true, will only keep the application context and its docker image.

in order to keep the same with the "--no-cache" option. name it as "--no-base".

run this cmd will build app image:`sealer build -f Kubefile -t myapp:v1 -m lite . --no-base`
if want keep all images include base image, run `sealer build -f Kubefile -t myimage:v1 -m lite .`

### Special notes for reviews
link to issue #695

### DIFF
--- a/build/buildkit/buildimage/build_image.go
+++ b/build/buildkit/buildimage/build_image.go
@@ -119,7 +119,7 @@ func (b BuildImage) genNewLayer(layerType, layerValue, filepath string) (v1.Laye
 	return imageLayer, nil
 }
 
-func (b BuildImage) SaveBuildImage(name string) error {
+func (b BuildImage) SaveBuildImage(name string, noBase bool) error {
 	cluster, err := b.getImageCluster()
 	if err != nil {
 		return err
@@ -129,7 +129,7 @@ func (b BuildImage) SaveBuildImage(name string) error {
 	if err != nil {
 		return fmt.Errorf("failed to set image metadata, err: %v", err)
 	}
-	layers, err := b.collectLayers()
+	layers, err := b.collectLayers(noBase)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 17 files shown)


build/buildkit/buildimage/build_image.go
```diff
@@ -119,7 +120,37 @@ func (b BuildImage) genNewLayer(layerType, layerValue, filepath string) (v1.Laye
 	return imageLayer, nil
 }
 
-func (b BuildImage) SaveBuildImage(name string) error {
+func (b BuildImage) checkImageMataData(opts SaveOpts) error {
```
fanux, line 123: Metadata

pkg/runtime/runtime.go
```diff
@@ -33,9 +33,10 @@ type Interface interface {
 }
 
 type Metadata struct {
-	Version string `json:"version"`
-	Arch    string `json:"arch"`
-	Variant string `json:"variant"`
+	Version     string `json:"version"`
+	Arch        string `json:"arch"`
+	Variant     string `json:"variant"`
+	KubeVersion string `json:"kubeVersion"`
```
fanux, line 39: What difference to Version?

sealer/cmd/build.go
```diff
@@ -74,7 +81,9 @@ func init() {
 	buildCmd.Flags().StringVarP(&buildConfig.BuildType, "mode", "m", "lite", "cluster image build type, default is lite")
 	buildCmd.Flags().StringVarP(&buildConfig.KubefileName, "kubefile", "f", "Kubefile", "kubefile filepath")
 	buildCmd.Flags().StringVarP(&buildConfig.ImageName, "imageName", "t", "", "cluster image name")
+	buildCmd.Flags().StringVarP(&buildConfig.KubeVersion, "kubeversion", "k", "", "KubeVersion is a SemVer constraint specifying the version of Kubernetes required.")
```
fanux, line 84: Why needs this argument?

pkg/runtime/runtime.go
kakaZhou719: use KubeVersion to save version constraint.

sealer/cmd/build.go
kakaZhou719: i think user could specify the version constraint

fanux:
I think `--base false` is better, default is true.

fanux (approved):

PR #820, opened 2021-11-03
Upload all contents under binary folder

Stevent-fei:


fanux:
Also need add the ossurl into release page.

fanux (approved):

Issue #814, opened 2021-11-02
sealer run support specific infra provider


kakaZhou719:
## Issue Description

we use default clusterfile form cloud image manifest to execute  sealer run , while if user want overwrite the provider ,such as container ,or another cloud.

Type: *feature request*

bxy4543:
#822

PR #822, opened 2021-11-04
Feature/support run provider

bxy4543:
1.  sealer run support specific infra provider #814 
2. fix kubeadm template error caused by ip port

### DIFF
--- a/apply/run.go
+++ b/apply/run.go
@@ -29,16 +29,9 @@ import (
 )
 
 type ClusterArgs struct {
-	cluster    *v1.Cluster
-	imageName  string
-	nodeArgs   string
-	masterArgs string
-	user       string
-	passwd     string
-	pk         string
-	pkPasswd   string
-	podCidr    string
-	svcCidr    string
+	cluster   *v1.Cluster
+	imageName string
+	runArgs   *common.RunArgs
 }
 
 func IsNumber(args string) bool {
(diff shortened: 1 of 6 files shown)


sealer/cmd/run.go
```diff
@@ -52,6 +55,7 @@ create cluster to your baremetal server, appoint the iplist:
 func init() {
 	runArgs = &common.RunArgs{}
 	rootCmd.AddCommand(runCmd)
+	runCmd.Flags().StringVarP(&runArgs.Provider, "provider", "", "", "set infra provider, example `ALI_CLOUD`, the local server need ignore this")
```
kakaZhou719, line 58: short args: "-p" ?

apply/run.go
```diff
@@ -83,35 +76,40 @@ func (c *ClusterArgs) SetClusterArgs() error {
...
 	}
-	if IsNumber(c.masterArgs) && (IsNumber(c.nodeArgs) || c.nodeArgs == "") {
-		c.cluster.Spec.Masters.Count = c.masterArgs
-		c.cluster.Spec.Nodes.Count = c.nodeArgs
-		c.cluster.Spec.Provider = common.DefaultCloudProvider
-	} else if IsIPList(c.masterArgs) && (IsIPList(c.nodeArgs) || c.nodeArgs == "") {
-		c.cluster.Spec.Masters.IPList = strings.Split(c.masterArgs, ",")
-		if c.nodeArgs != "" {
-			c.cluster.Spec.Nodes.IPList = strings.Split(c.nodeArgs, ",")
+	if IsNumber(c.runArgs.Masters) && (IsNumber(c.runArgs.Nodes) || c.runArgs.Nodes == "") {
```
kakaZhou719, line 95: if utils.InList(c.runArgs.Provider, []string{common.AliCloud, common.CONTAINER}) {
c.cluster.Spec.Provider = c.runArgs.Provider
}
i think here we don’t  need to double check the value of c.runArgs.Provider

sealer/cmd/run.go
bxy4543: `passwd` already use `-p`

apply/run.go
bxy4543: If the provider is a Cloud mod but the masters or Nodes type is set to ipList,  i think need to check this.

bxy4543: If the provider is a Cloud mod but the masters or Nodes type is set to ipList,  i think need to check this.

fanux:
Maybe we can use the cobra validate function to validate args.

fanux (approved):

Issue #265, opened 2021-06-05
support rmi with partial id


touchuyht:
Type: feature request
Currently, we only support rmi with full length id of a image which is not flexible. I suggest we could remove images with just partial id of the image.

Describe what feature you want
[root@handsome ~]# sealer images
+----------------------------------+----------------------------------------------------------------------+
|             IMAGE ID             |                              IMAGE NAME                              |
+----------------------------------+----------------------------------------------------------------------+
| 8233ad647e35d6ec32d98efe45467cb5 | registry.cn-hongkong.aliyuncs.com/sealyun/kubernetes-amd64-1:v1.19.9 |
| 8233ad647e35d6ec32d98efe45467cb5 | registry.cn-hongkong.aliyuncs.com/sealyun/kubernetes-amd64:v1.19.9   |
sealer rmi 8233 -f

Additional context

PR #824, opened 2021-11-04
support rmi with partial id

bxy4543:
support rmi with partial id #265

### DIFF
--- a/image/default_image.go
+++ b/image/default_image.go
@@ -18,24 +18,21 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"strings"
 
+	"github.com/alibaba/sealer/common"
+	"github.com/alibaba/sealer/image/distributionutil"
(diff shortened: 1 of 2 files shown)


image/default_image.go
```diff
@@ -170,50 +172,59 @@ func (d DefaultImageService) Login(RegistryURL, RegistryUsername, RegistryPasswd
...
 
-func (d DefaultImageService) Delete(imageName string) error {
+func (d DefaultImageService) Delete(imageArg string) error {
 	var (
 		images        []*v1.Image
 		image         *v1.Image
 		imageTagCount int
 		imageID       string
 		imageStore    = d.imageStore
+		isName        = strings.Contains(imageArg, ":")
```
fanux, line 182: Using this to judge a args is name or not is not very good.

You can just try to delete it if it is a name, then try to delete it as ID

fanux (approved):

PR #877, opened 2021-11-18
Bugfix apply upgrade

kakaZhou719:
### Describe how you did it
1. create : delete "preinstall plugin", no plugin  phase is "preinstall ", if we want to pre-config hosts, we can use "preinit"
2.  scale : delete "guest apply ",no need to run guest apply again when scale up.
3.  upgrade :delete "guest apply ",no need to run guest apply again when upgrade .
4.  upgrade : use k8sClient to fetch current cluster version.


shy-Xu:
good job

Issue #705, opened 2021-09-24
docker precheck


fanux:
Precheck docker version is sealer-docker.

Sealer not support other docker versions now, because sealer needs cache all docker image.

zsyaoo:
take it

fanux:
https://github.com/alibaba/sealer/pull/893/files @zsyaoo use kyverno sealer support raw docker now, so I think we may not need this checker anymore.

PR #878, opened 2021-11-18
add container checker.

zsyaoo:
#705 add container checker


fanux:
https://github.com/alibaba/sealer/pull/893/files @zsyaoo use kyverno sealer support raw docker now, so I think we may not need this checker anymore.

Issue #865, opened 2021-11-17
Store the required docker image in private registry when execute "sealer build"


shy-Xu:
## Issue Description

Type: *feature request*

### Describe what feature you want
we need to build the users required docker image into base image when execute "sealer build",then users should only pull images from private registry embedded in base image when they use sealer to create a cluster.

preliminary workflow : when execute "sealer build" ,use docker pull to pull required images from outside public registry,then push those images to our pirvate registry.

shy-Xu:
this feature is related to issue #859

PR #881, opened 2021-11-18
Build can store docker image to private registry when cluster uses official raw docker

shy-Xu:
this PR is aimed to solve the issue #865

### DIFF
--- a/build/buildkit/buildimage/build_image.go
+++ b/build/buildkit/buildimage/build_image.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/alibaba/sealer/build/buildkit/buildinstruction"
 	"github.com/alibaba/sealer/client/docker"
@@ -34,7 +35,8 @@ import (
 )
 
 const (
-	maxLayerDeep = 128
+	maxLayerDeep           = 128
+	sealerDockerServerFlag = `sealer`
 )
 
 // BuildImage this struct aims to provide image object in build stage
@@ -62,7 +64,16 @@ func (b BuildImage) ExecBuild(ctx Context) error {
 	} else {
 		execCtx = buildinstruction.NewExecContextWithoutCache(ctx.BuildType, ctx.BuildContext, b.LayerStore)
 	}
-	execCtx.RawDocker = ctx.RawDocker
+	//judge the docker engine version
+	dockerServerVersion, err := b.DockerClient.GetServerVersion()
+	if err != nil {
+		return err
+	}
+	if strings.Contains(dockerServerVersion, sealerDockerServerFlag) {
+		execCtx.SealerDocker = true
+	} else {
+		execCtx.SealerDocker = false
+	} //judge the docker engine version over
 	for i := 0; i < len(newLayers); i++ {
 		//we are to set layer id for each new layers.
 		layer := &newLayers[i]
--- a/build/buildkit/buildimage/build_image.go
+++ b/build/buildkit/buildimage/build_image.go
@@ -62,6 +62,7 @@ func (b BuildImage) ExecBuild(ctx Context) error {
 	} else {
 		execCtx = buildinstruction.NewExecContextWithoutCache(ctx.BuildType, ctx.BuildContext, b.LayerStore)
 	}
+	execCtx.RawDocker = ctx.RawDocker
 	for i := 0; i < len(newLayers); i++ {
 		//we are to set layer id for each new layers.
 		layer := &newLayers[i]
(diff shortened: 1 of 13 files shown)


shy-Xu:
After discusstion with @kakaZhou719 ,I design a way to solve issue #865.Now,I add a flag to identify whether the cluster use raw docker.I also complete the interface definations and functions call.I will finish the implement of these functions later.

shy-Xu:
I have completed the code development, but the check is failed for unknow reason

client/docker/images.go
```diff
@@ -29,32 +29,92 @@ import (
...
+		if strings.HasPrefix(image, "#") {
+			continue
+		}
+		if err := d.ImagePull(strings.TrimSpace(image)); err != nil {
+			return fmt.Errorf("image %s pull failed: %v", image, err)
+		}
+	}
+	return nil
+}
+
```
shy-Xu, line 106: I comment out these two functions because I find they are not used

shy-Xu:
Now users need not input a flag to identify whether it's sealer docker or raw docker

build/buildkit/buildimage/build_image.go
```diff
@@ -62,6 +64,16 @@ func (b BuildImage) ExecBuild(ctx Context) error {
 	} else {
 		execCtx = buildinstruction.NewExecContextWithoutCache(ctx.BuildType, ctx.BuildContext, b.LayerStore)
 	}
+	//judge the docker engine version
+	dockerServerVersion, err := b.DockerClient.GetServerVersion()
```
fanux, line 68: Check the version in `ImagesCacheToRegistry`, so you don't need the bool argument

client/docker/images.go
```diff
@@ -29,32 +29,96 @@ import (
...
+	err = d.cli.ImageTag(d.ctx, named.String(), TagImageName)
+	if err != nil {
+		return err
+	}
+	out, err = d.cli.ImagePush(d.ctx, TagImageName, opts)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = out.Close()
```
fanux, line 86: should not ignore errors

client/docker/images.go
```diff
@@ -29,32 +29,96 @@ import (
...
 	if err != nil {
-		logger.Error(fmt.Sprintf("Read image list failed: %v", err))
+		return fmt.Errorf("failed to parse canonical image name %s : %v", image, err)
 	}
-	return d.ImagesPull(data)
-}
+	TagImageName = strings.Replace(named.String(), named.String()[0:strings.Index(named.String(), `/`)], registryName, 1)
+	opts := types.ImagePushOptions(GetCanonicalImagePullOptions(named.String()))
+	if opts.RegistryAuth == "" {
+		opts.RegistryAuth = "sea.hub" //must not empty
```
fanux, line 75: Use const

client/docker/images.go
```diff
@@ -29,32 +29,96 @@ import (
 	dockerjsonmessage "github.com/docker/docker/pkg/jsonmessage"
 )
 
-func (d Docker) ImagesPull(images []string) error {
+func (d Docker) ImagesCacheToRegistry(images []string) error {
+	version, err := d.cli.ServerVersion(d.ctx)
+	if err != nil {
+		return err
+	}
+	if strings.Contains(version.Version, `sealer`) {
```
fanux, line 37: const

client/docker/images.go
```diff
@@ -29,32 +29,96 @@ import (
...
+		TagImageName string
+	)
+	named, err = GetCanonicalImageName(image)
 	if err != nil {
-		logger.Error(fmt.Sprintf("Read image list failed: %v", err))
+		return fmt.Errorf("failed to parse canonical image name %s : %v", image, err)
 	}
-	return d.ImagesPull(data)
-}
+	TagImageName = strings.Replace(named.String(), named.String()[0:strings.Index(named.String(), `/`)], registryName, 1)
```
fanux, line 72: How about the image name don't contains `/` ?

shy-Xu: named.String() returns a image name which has been handled. For example, if the input image name is "ubuntu", named.String() returns docker.io/library/ubuntu:latest

fanux (approved):

client/docker/images.go
```diff
@@ -29,6 +29,27 @@ import (
...
 )
 
+const (
+	sealerDockerString = `sealer`
+	registryAuthString = `sea.hub`
+	registryName       = `sea.hub:5000`
+)
+
+func (d Docker) ImagesCacheToRegistry(images []string) error {
+	version, err := d.cli.ServerVersion(d.ctx)
```
kakaZhou719, line 39: whether is sealer-docker or not, we can move out this as an utils flag. And in some scenarios we don’t even have docker, we can try to use cmd "docker version | grep sealer".

PR #889, opened 2021-11-22
cluster v2 Check module

shy-Xu:
this PR is for solving issue#872

### DIFF
--- a/pkg/checker/host_checker.go
+++ b/pkg/checker/host_checker.go
@@ -19,7 +19,6 @@ import (
 	"strconv"
 	"time"
 
-	v1 "github.com/alibaba/sealer/types/api/v1"
 	v2 "github.com/alibaba/sealer/types/api/v2"
 
 	"github.com/alibaba/sealer/utils/ssh"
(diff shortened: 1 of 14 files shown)


shy-Xu:
checker module has a function in file host_checker.go,this func calls another func included in package ssh.But package ssh has not upgraded to cluster v2,so there is a error exist.

fanux:
Not modify the v1 code, add v2 into pkg/checker, we delete v1 when all test access.

pkg/checker/host_checker.go
```diff
@@ -0,0 +1,92 @@
...
+	v2 "github.com/alibaba/sealer/types/api/v2"
+
+	"github.com/alibaba/sealer/utils/ssh"
+)
+
+type HostChecker struct {
+}
+
+func (a HostChecker) Check(cluster *v2.Cluster, phase string) error {
+	ssh := ssh.NewSSHClient(&cluster.Spec.SSH)
```
fanux, line 32: Each node may has it own ssh config, checkout this: https://github.com/alibaba/sealer/blob/main/pkg/runtime/kubeadm_runtime.go#L94

1. move the function to ssh module,  GetHostSSHClientByIP(hostIP string, *v2.Cluster)
2. The function has bug, it don't merge the ssh config when host ssh config not exist.

fanux: I will fix those issue.

pkg/checker/host_checker.go
```diff
@@ -0,0 +1,86 @@
...
+func NewHostChecker() Interface {
+	return &HostChecker{}
+}
+
+func checkHostnameUnique(cluster *v2.Cluster, ipList []string) error {
+	hostnameList := map[string]bool{}
+	for _, ip := range ipList {
+		s, err := ssh.GetHostSSHClient(ip, cluster)
+		if err != nil {
+			return err
```
fanux, line 51: wrap error.

fanux (approved):

PR #894, opened 2021-11-24
add V2 SSH Interface for V2

bxy4543:
add V2 SSH Interface for V2

### DIFF
--- a/utils/ssh/v2_ssh.go
+++ b/utils/ssh/v2_ssh.go
@@ -22,11 +22,12 @@ import (
 
 	"github.com/alibaba/sealer/common"
 	"github.com/alibaba/sealer/logger"
-	v1 "github.com/alibaba/sealer/types/api/v1"
+	v2 "github.com/alibaba/sealer/types/api/v2"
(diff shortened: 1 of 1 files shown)


utils/ssh/v2_ssh.go
```diff
@@ -0,0 +1,139 @@
...
+	"time"
+
+	"github.com/alibaba/sealer/common"
+	"github.com/alibaba/sealer/logger"
+	v2 "github.com/alibaba/sealer/types/api/v2"
+	"github.com/alibaba/sealer/utils"
+	"github.com/imdario/mergo"
+)
+
+type V2Interface interface {
```
fanux, line 30: Should we need change the ssh interface?  I think we just need add a New ssh function: https://github.com/alibaba/sealer/pull/891/files

fanux: Then new a ssh client by IP and v2.Cluster: https://github.com/alibaba/sealer/pull/882/files#diff-22657dc896bfef614095b0434316b07105c9d66099cb05b60ccbfed54c3fcb8cR94

PR #912, opened 2021-12-03
clean code and fix remote connection leak

wenjun93:
correct error handling, connection leak fixing, remove redundent code, etc.


utils/ssh/connect.go
```diff
@@ -65,10 +65,7 @@ func (s *SSH) Connect(host string) (*ssh.Client, *ssh.Session, error) {
 
 	session, err := client.NewSession()
 	if err != nil {
-		err := client.Close()
```
fanux, line 68: Why ignore the error?

wenjun93: @fanux I think the top level error should be returned instead

bxy4543 (approved):

PR #923, opened 2021-12-08
doc : Support build ARM CloudImage in AMD environment

kakaZhou719:


fanux:
Using ImageSave, so needn't qemu anymore

PR #941, opened 2021-12-16
fix upgrade get k8s version

bxy4543:
fix apply upgrade to get kubernetes version from kubeadm config instead of Metadata.
encapsulate phase exec plugin function

### DIFF
--- a/apply/v2/applydriver/local.go
+++ b/apply/v2/applydriver/local.go
@@ -16,7 +16,6 @@ package applydriver
 
 import (
 	"fmt"
-	"path/filepath"
 
 	v2 "github.com/alibaba/sealer/types/api/v2"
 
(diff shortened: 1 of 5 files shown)


apply/v2/applydriver/local.go
```diff
@@ -165,22 +166,30 @@ func (c *Applier) upgradeCluster(mj, nj []string) error {
...
 	}
 	// fetch form exec machine
-	desiredMetadata, err := runtime.LoadMetadata(filepath.Join(common.DefaultMountCloudImageDir(c.ClusterDesired.Name),
-		common.DefaultMetadataName))
+	rt, err := runtime.NewDefaultRuntime(c.ClusterDesired, c.ClusterDesired.GetAnnotationsByKey(common.ClusterfileName))
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to init runtime, %v", err)
+	}
+	KubeadmConfig := &runtime.KubeadmConfig{}
```
fanux, line 173: Apply module should not care about kubeadmConfig, what if runtime is k3s? You can add a function in Interface like GetClusterMetadata.

pkg/runtime/runtime.go
```diff
@@ -29,6 +29,7 @@ type Interface interface {
 	JoinNodes(newNodesIPList []string) error
 	DeleteMasters(mastersIPList []string) error
 	DeleteNodes(nodesIPList []string) error
+	GetClusterMetadata(cluster *v2.Cluster) (*KubeadmConfig, error)
```
fanux, line 32: should not return KubeadmConfig, Other runtime don't have kubeadm..

fanux (approved):

PR #954, opened 2021-12-22
Add the -e parameter and call wrapper

Stevent-fei:

### DIFF
--- a/common/structs.go
+++ b/common/structs.go
@@ -24,4 +24,5 @@ type RunArgs struct {
 	PodCidr    string
 	SvcCidr    string
 	Provider   string
+	CustomEnv  string
 }
--- a/common/structs.go
+++ b/common/structs.go
@@ -24,5 +24,5 @@ type RunArgs struct {
 	PodCidr    string
 	SvcCidr    string
 	Provider   string
-	CustomEnv  string
+	CustomEnv  []string
 }
--- a/pkg/plugin/shell_plugin.go
+++ b/pkg/plugin/shell_plugin.go
@@ -16,6 +16,7 @@ package plugin
 
 import (
 	"fmt"
+	"github.com/alibaba/sealer/pkg/env"
 	"strings"
 
 	"github.com/alibaba/sealer/utils"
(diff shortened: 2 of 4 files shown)


sealer/cmd/run.go
```diff
@@ -68,6 +68,7 @@ func init() {
 	runCmd.Flags().StringVarP(&runArgs.PkPassword, "pk-passwd", "", "", "set baremetal server  private key password")
 	runCmd.Flags().StringVarP(&runArgs.PodCidr, "podcidr", "", "", "set default pod CIDR network. example '10.233.0.0/18'")
 	runCmd.Flags().StringVarP(&runArgs.SvcCidr, "svccidr", "", "", "set default service CIDR network. example '10.233.64.0/18'")
+	runCmd.Flags().StringVarP(&runArgs.CustomEnv, "customEnv", "e", "", "set custom environment variables. example 'sealer run -e DashBoardPort=8443 mydashboard:latest'")
```
bxy4543, line 71: string type is replaced with the slice type（runCmd.Flags().StringSliceVarP）

runArgs.CustomEnv needs to be written to the cluster in `NewApplierFromArgs(args[0], runArgs)` function

pkg/plugin/shell_plugin.go
```diff
@@ -80,3 +83,20 @@ func (s Sheller) Run(context Context, phase Phase) error {
 func init() {
 	Register(ShellPlugin, &Sheller{})
 }
+
+func CustomEnv(context Context) (env.Interface, error) {
```
fanux, line 87: Where do you call this function, and why return the env.Interface.

common/structs.go
```diff
@@ -24,4 +24,5 @@ type RunArgs struct {
 	PodCidr    string
 	SvcCidr    string
 	Provider   string
+	CustomEnv  []string
```
kakaZhou719, line 27: i think the arg `CustomEnv ` should be `map[string]string`  type. why we use `[]string`.

kakaZhou719, line 27: I’m curious about why this structure `RunArgs` defined in common? it should in apply module.

Stevent-fei: My understanding is that the execution command, for example, sealer run image -e port=6644 -m xxx -n xxx -p xxx should be in the run parameter, and the run parameter is in the common module.

fanux (approved):

Issue #952, opened 2021-12-21
sealer exec


fanux:
exec a shell command or script on all node.

```
sealer exec `cat /etc/hosts`
```

bxy4543:
#956

PR #956, opened 2021-12-22
#952 feature: add sealer exec command

zsyaoo:
#952 feature: add sealer exec command

### DIFF
--- a/sealer/cmd/exec.go
+++ b/sealer/cmd/exec.go
@@ -1,0 +1,65 @@
+// Copyright © 2021 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"github.com/alibaba/sealer/common"
+	"github.com/alibaba/sealer/utils"
+	"github.com/alibaba/sealer/utils/ssh"
+	"github.com/spf13/cobra"
+)
+
+// execCmd represents the exec command
+var execCmd = &cobra.Command{
+	Use:   "exec",
+	Short: "exec a shell command or script on all node.",
+	Example: `
+exec to default cluster: my-cluster
+	sealer exec cat /etc/hosts
+specify the cluster name(If there is only one cluster in the $HOME/.sealer directory, it should be applied. ):
+    sealer exec -c my-cluster cat /etc/hosts
+`,
+	Args: cobra.NoArgs,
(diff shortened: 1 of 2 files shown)


sealer/cmd/exec.go
```diff
@@ -0,0 +1,75 @@
...
+		cluster, err := utils.GetClusterFromFile(clusterFile)
+		if err != nil {
+			return err
+		}
+		ipList := append(cluster.GetMasterIPList(), cluster.GetNodeIPList()...)
+		var wg sync.WaitGroup
+		for _, ip := range ipList {
+			sshCli, err := ssh.GetHostSSHClient(ip, cluster)
+			if err != nil {
+				return err
```
fanux, line 56:
```
	Google := func(ctx context.Context, query string) ([]Result, error) {
		g, ctx := errgroup.WithContext(ctx)

		searches := []Search{Web, Image, Video}
		results := make([]Result, len(searches))
		for i, search := range searches {
			i, search := i, search // https://golang.org/doc/faq#closures_and_goroutines
			g.Go(func() error {
				result, err := search(ctx, query)
				if err == nil {
					results[i] = result
				}
				return err
			})
		}
		if err := g.Wait(); err != nil {
			return nil, err
		}
		return results, nil
	}
```

Using errgroup https://pkg.go.dev/golang.org/x/sync/errgroup.

And you can put then into a module named exec.

sealer/cmd/exec.go
```diff
@@ -0,0 +1,50 @@
...
+specify the cluster name(If there is only one cluster in the $HOME/.sealer directory, it should be applied. ):
+    sealer exec -c my-cluster 'cat /etc/hosts'
+set role label to exec cmd:
+    sealer exec -c my-cluster -r master,slave,node1 'cat /etc/hosts'		
+`,
+	Args: cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		execCmd, err := exec.NewExecCmd(clusterName, roles)
+		if err != nil {
+			return nil
```
fanux, line 40: return err

sealer/cmd/exec.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+		return execCmd.RunCmd(args...)
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(execCmd)
+	execCmd.Flags().StringVarP(&clusterName, "cluster-name", "c", "", "submit one cluster name")
+	execCmd.Flags().StringVarP(&roles, "roles", "r", "", "set role label to roles")
```
fanux, line 49: roles should be string slice.

zsyaoo: we can use  `sealer exec -c my-cluster -r master,node,test1 'cat /etc/hosts'  `

sealer/cmd/exec.go
zsyaoo: i fixed

sealer/cmd/exec.go
zsyaoo: > roles should be string slice.
we can use sealer exec -c my-cluster -r master,node,test1 'cat /etc/hosts'

zsyaoo: we can use `sealer exec -c my-cluster -r master,node,test1 'cat /etc/hosts'`

fanux (approved):

PR #976, opened 2021-12-27
fix a bug about image save

shy-Xu:
This PR solve a problem about image save module. However, it is unsafe so the "golang-security-action" is failed.

what's unsafe is that it calls http.Get (in distribution/distribution source code) with a variable argument instead of a const argument.

### DIFF
--- a/image/save/distributionpkg/proxy/proxyauth.go
+++ b/image/save/distributionpkg/proxy/proxyauth.go
@@ -86,6 +86,7 @@ func getAuthURLs(remoteURL string) ([]string, error) {
 	return authURLs, nil
 }
 
+// #nosec
 func ping(manager challenge.Manager, endpoint string) error {
 	resp, err := http.Get(endpoint)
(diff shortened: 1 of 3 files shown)


image/save/distributionpkg/proxy/proxyauth.go
```diff
@@ -82,3 +84,13 @@ func getAuthURLs(remoteURL string) ([]string, error) {
 
 	return authURLs, nil
 }
+
+func ping(manager challenge.Manager, endpoint, versionHeader string) error {
+	resp, err := http.Get(endpoint)
```
shy-Xu, line 89: Here is unsafe

bxy4543: you can skip the security action with '// #nosec',
or will it a safe way, reference: https://securego.io/docs/rules/g107.html

fanux (approved):

PR #984, opened 2021-12-29
update docs about kyverno

shy-Xu:
This PR update some contents about kyverno and raw docker documents


fanux:
I think the docs title should be `use kyverno base image`, not kyverno build-in

fanux (approved):

PR #992, opened 2021-12-30
fix apply error when env is empty,

bxy4543:
fix apply error when env is empty,
fix plugin out of tree err 'plugin type not registered'
fix compat v1 cluster env


plugin/plugins.go
```diff
@@ -92,6 +92,9 @@ func (c *PluginsProcessor) Load() error {
 // Run execute each in-tree or out-of-tree plugin by traversing the plugin list.
 func (c *PluginsProcessor) Run(cluster *v1.Cluster, phase Phase) error {
 	for _, config := range c.Plugins {
+		if config.Spec.Action != string(phase) {
```
kakaZhou719, line 95: package plugin was moved to pkg .this will be deprecated

Stevent-fei (approved):

PR #993, opened 2021-12-30
optimize v2 test

bxy4543:
v2 Infra module has not been upgraded, and the test also temporarily shields infra test;
Because infra has not implemented V2 cluster to pull infra, and v2 Apply is compatible with v1, the test will not be upgraded to v2 for the time being.
Add debug mode to display debug logs for e2e tests.

### DIFF
--- a/test/suites/apply/apply.go
+++ b/test/suites/apply/apply.go
@@ -85,6 +85,7 @@ func LoadPluginFromDisk(clusterFilePath string) []v1.Plugin {
 func GenerateClusterfile(clusterfile string) {
 	filepath := GetRawConfigPluginFilePath()
 	cluster := LoadClusterFileFromDisk(clusterfile)
+	cluster.Spec.Env = []string{"env=TestEnv"}
 	data, err := yaml.Marshal(cluster)
 	testhelper.CheckErr(err)
 	appendData := [][]byte{data}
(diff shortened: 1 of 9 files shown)


test/suites/build/fixtures/cloud_build/Kubefile
```diff
@@ -1,10 +1,10 @@
-FROM sealer-io/test:v1
+FROM sealer-io/kubernetes:v1.19.8
```
kakaZhou719, line 1: cloud build will push image automatically ,this action will update our base image ,so i suggested that we use the test base image.

test/suites/build/fixtures/cloud_build/imageList
```diff
@@ -0,0 +1,4 @@
+quay.io/tigera/operator:v1.17.4
+docker.io/redis:6.2.5
+registry.cn-qingdao.aliyuncs.com/sealer-apps/debug:ubuntu
```
kakaZhou719, line 3: less docker image we add ,less time we cost

test/suites/apply/apply.go
```diff
@@ -122,11 +123,11 @@ func GenerateClusterfile(clusterfile string) {
 }
 
 func SealerDeleteCmd(clusterFile string) string {
-	return fmt.Sprintf("%s delete -f %s --force", settings.DefaultSealerBin, clusterFile)
+	return fmt.Sprintf("%s delete -f %s --force -d", settings.DefaultSealerBin, clusterFile)
```
fanux, line 126: What is -d for ?

kakaZhou719 (approved):

PR #1001, opened 2021-12-30
support save helm chart

### LINKED ISSUE #966: sealer can pull helm charts when sealer build
## Issue Description
Because helm charts are different from docker images in terms of structure, docker registry codes failed to pull helm charts. We need find out how helm pulls charts from helm repository, then integrate this ability to sealer.

There are two problems need to solve:
First, how to store charts? Two ways, run another charts registry or modify current docker registry to make it compatible with helm charts. I prefer latter.

Second, how to pull charts? I think it's better to redefine interface:

```
//Image interface can save a list of images of the specified platform
type Image interface {
	SaveImages(images []string, platform v1.Platform) error
}

type chart struct {
	name string //e.g: mysql
	repo string // e.g: https://charts.bitnami.com/bitnami
}

// Chart interface can save a list of helm charts
type Chart interface {
	SaveCharts(charts []chart) error
}

//ChartImage interface can save both docker images and helm charts
type ChartImage interface {
	Image
	Chart
}

//DefaultSaver implement ChartImage interface
type DefaultSaver struct {
	ctx            context.Context
	rootdir        string
	domainToImages map[string][]Named
	progressOut    progress.Output
}

//NewSaver receive two arguments and create a DefaultSaver
func NewSaver(ctx context.Context, rootdir string) ChartImage {
	if ctx == nil {
		ctx = context.Background()
	}
	return &DefaultImageSaver{
		ctx:            ctx,
		rootdir:        rootdir,
		domainToImages: make(map[string][]Named),
	}
}
```

shy-Xu:
Fixes #966


shy-Xu:
This is preliminary work, finish store helm chart.

I will improve it latter, e.g: error handles

image/save/chartsave.go
```diff
@@ -0,0 +1,97 @@
...
+	"path/filepath"
+
+	"github.com/alibaba/sealer/common"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/repo"
+)
+
+const (
+	CHARTDIR = "/charts"
```
fanux, line 30: Why save charts to this dir, registry will not mount this dir

Issue #1026, opened 2022-01-05
Can give a default image name or tag for sealer merged  image


mahmut-Abi:
## Issue Description

Type: *feature request*

### Describe what feature you want
after sealer merge without -t name:tag there is no name and  tag for new merged image.
after sealer merge with -t name but without tag, there is only image name whitout tag, can we give it a default tag like 'latest'

mahmut-Abi:
![image](https://user-images.githubusercontent.com/96514873/148316399-72e257b4-b249-474a-bbf4-fd70e6c20680.png)
the image whiout tag can not be delete.

PR #1035, opened 2022-01-06
Default tag

mahmut-Abi:
add the default name (merged) and tag(latest) to sealer merged name.

#1026


sealer/cmd/merge.go
```diff
@@ -43,6 +43,12 @@ merge images:
 			}
 			images = append(images, image)
 		}
+		if ImageName == "" {
+			ImageName = "merged:latest"
```
fanux, line 47: I think ImageName is a required argument.

Issue #1047, opened 2022-01-07
failed to run sealer check --pre


kakaZhou719:
#### What happened:
`sealer check --pre `and `sealer check --post` have the same behavior logic

![image](https://user-images.githubusercontent.com/83740799/148481776-b7cee350-da93-4bcc-9e1c-8de5ca962fd0.png)

#### How to reproduce it (as minimally and precisely as possible):

![image](https://user-images.githubusercontent.com/83740799/148481700-0381c729-616d-4f24-993c-04ac0f60d6e9.png)

#### Environment:

- sealer version (use `sealer version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #1063, opened 2022-01-10
fix run sealer check --pre failed

bxy4543:
fix run sealer check --pre failed #1047

### DIFF
--- a/pkg/checker/check_list.go
+++ b/pkg/checker/check_list.go
@@ -23,18 +23,11 @@ import (
 const (
 	PhasePre  = "Pre"
 	PhasePost = "Post"
-	PhaseView = "view"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/checker/check_list.go
```diff
@@ -21,22 +21,16 @@ import (
 )
 
 const (
-	PhasePre  = "Pre"
-	PhasePost = "Post"
-	PhaseView = "view"
+	PhasePre      = "Pre"
+	PhasePost     = "Post"
+	PhasePostView = "PostView"
```
fanux, line 26: What does post view means?

bxy4543: by default post check does not display information, post view means post phase and view more info

fanux (approved):

Issue #1048, opened 2022-01-07
whats the main propose of sealer exec ?


kakaZhou719:
#### What happened:

follow the example, the command line result is not echoed.whats the main propose of sealer exec ? I understand it should be able to execute commands remotely，and show the result.

![image](https://user-images.githubusercontent.com/83740799/148481986-78d220dd-da5c-489d-9b9c-8907624466b5.png)

#### Environment:

- sealer version (use `sealer version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

bxy4543:
The command return information has been optimized, so the --debug parameter is required to display information

fanux:
Should not using debug by default, should open the ssh.CmdAsync stdoutput.

PR #1077, opened 2022-01-12
fix:sealer exec not show cmd result

kakaZhou719:

lint to issue #1048

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -59,20 +59,20 @@ func NewExecCmd(clusterName string, roles string) (Exec, error) {
 	return Exec{cluster: cluster, ipList: ipList}, nil
 }
 
-func (exec *Exec) RunCmd(args ...string) error {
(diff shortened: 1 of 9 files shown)


pkg/exec/exec.go
```diff
@@ -59,20 +59,20 @@ func NewExecCmd(clusterName string, roles string) (Exec, error) {
...
+	for _, ipAddr := range e.ipList {
+		ip := ipAddr
 		eg.Go(func() error {
-			sshClient, sshErr := ssh.GetHostSSHClient(ip, exec.cluster)
+			sshClient, sshErr := ssh.GetHostSSHClient(ip, e.cluster)
 			if sshErr != nil {
 				return sshErr
 			}
-			err := sshClient.CmdAsync(ip, args...)
+			bytes, err := sshClient.Cmd(ip, cmd)
```
fanux, line 71: CmdAsync is better, we can add a flag to sshClient `stdout true`.

fanux (approved):

PR #1087, opened 2022-01-13
check clean.sh file is exists and add rm /usr/bin/kube* when delete c…

### LINKED ISSUE #1050: Sealer delete node needs to be stable enough to be as error-free as possible
sealer delete may also fail when executed after an installation failure, so sealer delete needs to be stable enough.

### LINKED ISSUE #1072: After deleting the cluster,/usr/bin/kubectl is not deleted
image

neo502721:
strengthen sealer delete process:

- skip execute clean.sh if it not exists 
- merge `rm /usr/bin/kubeadm` and `rm /usr/bin/kubelet` to `rm /usr/bin/kube*` in the RemoteCleanMasterOrNode.  This will also  delete /usr/bin/kubectl .

Fixes #1050 Fixes #1072 

### Describe how to verify it
sealer delete --all

### DIFF
--- a/pkg/filesystem/filesystem.go
+++ b/pkg/filesystem/filesystem.go
@@ -233,7 +233,10 @@ func unmountRootfs(ipList []string, cluster *v2.Cluster) error {
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


neo502721:
@bxy4543

fanux (approved):

PR #1097, opened 2022-01-16
fix a nil pointer bug

shy-Xu:
lQLPDhsOBpXrWXLNBATNBhSwd108D3_o19MB56clr0CGAA_1556_1028

### DIFF
--- a/utils/ssh/scp.go
+++ b/utils/ssh/scp.go
@@ -15,7 +15,6 @@
 package ssh
 
 import (
-	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


utils/ssh/scp.go
```diff
@@ -60,7 +60,7 @@ type easyProgressUtil struct {
 }
 
 //must call DisplayInit first
-func RegisterEpu(ip string, total int) {
+func registerEpu(ip string, total int) {
 	if progressChanOut == nil {
```
bxy4543, line 64: You still need to consider that the DisplayInit function is not called, so you can add init under the ssh.interface package to implement the DisplayInit function

bxy4543 (approved):

PR #1108, opened 2022-01-20
Build support add build args and render env in CMD

kakaZhou719:

1,add build arg from kubefile:

```
FROM kubernetes:v1.19.8
ARG name=testname
ARG PASSWORD=PASSWORD1
COPY nginx charts
RUN echo ${name} 
CMD echo ${PASSWORD}
```
`sealer build -t my-image:v1 . `
this will echo "testname".

2,add build arg from cmd line:

`sealer build -t my-image:v1 --build-arg PASSWORD=sealer123,name=sealer . `

this will echo "sealer".

3,use build arg at run stage:

`sealer run  my-image:v1 -e PASSWORD=sealer123 -m 192.168.1.1 -p ****`

this will echo "sealer123"  from CMD instruction.

priority order ： run cmd line > build cmd line > kubefile.


pkg/guest/guest.go
```diff
@@ -53,11 +57,27 @@ func (d *Default) Apply(cluster *v2.Cluster) error {
...
 			continue
 		}
-		if err := sshClient.CmdAsync(runtime.GetMaster0Ip(cluster), fmt.Sprintf(common.CdAndExecCmd, clusterRootfs, image.Spec.Layers[i].Value)); err != nil {
+
+		cmdline, err := ex.ProcessWordWithMap(image.Spec.Layers[i].Value, arg)
+		if err != nil {
+			return fmt.Errorf("failed to render build args: %v", err)
+		}
+
+		if err := sshClient.CmdAsync(runtime.GetMaster0Ip(cluster), fmt.Sprintf(common.CdAndExecCmd, clusterRootfs, cmdline)); err != nil {
```
bxy4543, line 80: `ex.ProcessWordWithMap(image.Spec.Layers[i].Value, arg)`
I think this `arg` should be cluster.env and support passing in parameters when sealer apply

kakaZhou719: yes, "arg" is the merged parameter from cluster.env and image  build-arg, and i already do some overwrite action.

pkg/parser/parser.go
```diff
@@ -131,9 +132,35 @@ func decodeLine(line string) (string, string, error) {
...
+	if len(valueLine) != 2 {
+		logger.Error("invalid ARG value %s. ARG format must be key=value", layerValue)
+		return
+	}
+	k := strings.TrimSpace(valueLine[0])
+	if !utils.IsLetterOrNumber(k) {
+		logger.Error("ARG key must be letter or number,invalid ARG format will ignore this key %s.", k)
+		return
+	}
+	ima.Spec.ImageConfig.Args[k] = strings.TrimSpace(valueLine[1])
```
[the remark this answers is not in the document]
kakaZhou719: good idea, but we need to keep the same user behavior to write the kubefile.

Stevent-fei (approved):

PR #1119, opened 2022-01-24
build arg instruction support newline list input

kakaZhou719:

below is my test kubefile: 
```
FROM kubernetes:v1.19.8
ARG value=sealer
ARG MY_ARG=abc
ARG MY_*ARG=abc # not support such key format.will ignoring this key.
ARG ddd= # default value is "".
ARG name=testname \
,PASSWORD1=PASSWORD1,\
aptget=update 
COPY aa .
COPY charts charts
RUN echo ${name}
RUN echo ${MY_ARG}
CMD echo ${PASSWORD1}
```

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -140,17 +140,21 @@ func dispatchArg(layerValue string, ima *v1.Image) {
 	if ima.Spec.ImageConfig.Args == nil {
 		ima.Spec.ImageConfig.Args = map[string]string{}
 	}
-	valueLine := strings.SplitN(layerValue, "=", 2)
-	if len(valueLine) != 2 {
-		logger.Error("invalid ARG value %s. ARG format must be key=value", layerValue)
-		return
+
+	kv := strings.Split(layerValue, ",")
+	for _, element := range kv {
+		valueLine := strings.SplitN(element, "=", 2)
+		if len(valueLine) != 2 {
+			logger.Error("invalid ARG value %s. ARG format must be key=value", layerValue)
+			return
+		}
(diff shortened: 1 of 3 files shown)


pkg/parser/parser.go
```diff
@@ -140,17 +140,21 @@ func dispatchArg(layerValue string, ima *v1.Image) {
...
-	}
-	k := strings.TrimSpace(valueLine[0])
-	if !utils.IsLetterOrNumber(k) {
-		logger.Error("ARG key must be letter or number,invalid ARG format will ignore this key %s.", k)
-		return
+
+	kv := strings.Split(layerValue, ",")
+	for _, element := range kv {
+		valueLine := strings.SplitN(element, "=", 2)
+		if len(valueLine) != 2 {
```
bxy4543, line 147: What if value has "="     
like: runArg="--masters=xxx --nodes=xxx"

kakaZhou719: if so, key="runArg" and value="--masters=xxx --nodes=xxx"。we only make sure that we got the the right key. the value is user set that is  meaningless for sealer build.

bxy4543 (approved):

fanux:
Add this feature into docs, and what diff with `ENV`?

kakaZhou719:
> Add this feature into docs, and what diff with `ENV`?

- build `ENV`  instruction: not support at present , it will add later for setting host temp env.
- build `ARG`  instruction: it will merge with `RUN` or `CMD` instruction line and  execute it .

more example usage see : https://github.com/alibaba/sealer/pull/1108

PR #1121, opened 2022-01-25
add route for multiple network interface

bxy4543:
add route for multiple network interface


pkg/runtime/nodes.go
```diff
@@ -123,6 +133,10 @@ func (k *KubeadmRuntime) deleteNodes(nodes []string) error {
 			if err := k.deleteNode(node); err != nil {
 				return fmt.Errorf("delete node %s failed %v", node, err)
 			}
+			err := k.deleteVIPRoute(node)
```
fanux, line 136: Why delete route here?

bxy4543: > Why delete route here?

after deleting a node, clear the traces of route addition.

pkg/runtime/nodes.go
```diff
@@ -38,6 +38,12 @@ const (
 	RemoteAddRoute                  = "seautil route add --host %s --gateway %s"
 	RouteOK                         = "ok"
 	LvscareStaticPodCmd             = `echo "%s" > %s`
+	CheckNetworkIsDefaultRoute      = "ip route show |grep default |grep %s"
+	DeleteRouteCmd                  = "if ip route |grep %s;then ip route del %s;fi"
+	AddRouteCmd                     = "ip route |grep %s || ip route add %s"
+	DeleteRouteFromFile             = "if [ -f /etc/sysconfig/network-scripts/route-%s ]; then sed -i \"/%s/d\" /etc/sysconfig/network-scripts/route-%s;fi"
+	AddStaticRouteFile              = "cat /etc/sysconfig/network-scripts/route-%s|grep %s || echo %s >> /etc/sysconfig/network-scripts/route-%s"
+	GetNetworkInterface             = "ip route |grep -v %s |grep %s| awk -F '[ \\t*]' '{print $3}'"
```
fanux, line 46: Should use ip addr command to get it. grep node ip

kakaZhou719: Some hosts do not have `ip` command, use "net.Interfaces()" to get network interface

kakaZhou719: Some hosts do not have `ip` command. 
if remote: "ifconfig" cmd is system build-in. 
if local : use "net.Interfaces()".

fanux: Or use sealutil to check interface and add router.

fanux (approved):

PR #1124, opened 2022-01-27
Add auto-build docs

Stevent-fei:
Add Github Action auto-build docs

### DIFF
--- a/docs/design/auto-build.md
+++ b/docs/design/auto-build.md
@@ -27,11 +27,11 @@ input:
 /imagebuild 1.20.14 amd64
 ```
 
-When `Image built successfully : kubernetes: version-auto` appears in the comment area, it indicates that the image build is completed.
+When `Image built successfully : kubernetes: version` appears in the comment area, it indicates that the image build is completed.
 
 Image name:
 
 ```shell
-kubernetes:v1.20.14-auto
+kubernetes:v1.20.14
 ```
 
--- a/docs/design/auto-build.md
+++ b/docs/design/auto-build.md
@@ -1,6 +1,6 @@
-# auto-build
+# Auto-build
 
-Automatic cluster image building is to meet the needs of some users. Users need to build a specified version of kubernetes for automatic building.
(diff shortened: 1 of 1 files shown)


fanux:
What the output CloudImage name? kubernetes:[version] like kubernetes:v1.19.8?

Stevent-fei:
> 

In order to distinguish between the basic image and the automatically built image, the suffix - Auto, for example: kubernetes: v1.19.8-auto

docs/design/auto-build.md
```diff
@@ -0,0 +1,40 @@
+# auto-build
```
kakaZhou719, line 1: upper case for title

docs/design/auto-build.md
```diff
@@ -0,0 +1,40 @@
...
+
+For example:
+
+```shell
+kubernetes:v1.20.14-auto
+```
+
+After inputting, GitHub action will start to execute the automatic construction of the basic image of the corresponding version of kubernetes.
+
+When `kubernetes: version-auto` appears in the comment area, it indicates that the image construction is completed.
```
kakaZhou719, line 42: use proper nouns： "image build" not "image construction"

bxy4543:
Since Auto-Build cannot be triggered by ordinary users, I think we should describe how to design it.

docs/design/auto-build.md
```diff
@@ -0,0 +1,37 @@
...
+```shell
+/imagebuild 1.20.14 amd64
+```
+
+When `Image built successfully : kubernetes: version-auto` appears in the comment area, it indicates that the image build is completed.
+
+Image name:
+
+```shell
+kubernetes:v1.20.14-auto
```
fanux, line 35: kubernetes:v1.20.14-auto -> kubernetes:v1.20.14

Stevent-fei: If the version number is changed directly, it will be confused with the basic image

fanux (approved):

PR #1129, opened 2022-01-28
Support cri image manager

### LINKED ISSUE #909: feat: improve image cache logic.
## Issue Description
kubeadm join --cri-socket /var/run/cri-resmgr/cri-resmgr.sock 
截屏2021-12-01 下午7 41 09

Sealer hacked docker for supporting image cache and request endpoint redirecting.
I wonder if the scenes like "docker pull xxx" are rare. Basically the image pulling would triggered by kubelet in sealer.
If the above assumption is for sure. I think we can proxy image pulling request like cri-resource-manager(https://github.com/intel/cri-resource-manager), which changes kubelet cri-socket to their own process, and make it easier for intercepting image pulling requests. And it doesn't have to change any logic of docker.

shy-Xu:
Fixes #909

### DIFF
--- a/pkg/runtime/nodes.go
+++ b/pkg/runtime/nodes.go
@@ -17,6 +17,7 @@ package runtime
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/alibaba/sealer/utils"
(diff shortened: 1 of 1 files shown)


shy-Xu:
The code add some handle when execute join nodes. To using this feature, you should use CloudImage `kubernetes-crishim:v1.19.8`

justadogistaken:
Maybe it's better to pass `image-service-endpoint` throgh `nodeRegistration.kubeletExtraArgs` [link](https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/) of kubeadm.

pkg/runtime/nodes.go
```diff
@@ -38,6 +39,8 @@ const (
 	RemoteAddRoute                  = "seautil route add --host %s --gateway %s"
 	RouteOK                         = "ok"
 	LvscareStaticPodCmd             = `echo "%s" > %s`
+	ReplaceImageSocket              = `sed '/KUBELET_EXTRA_ARGS/ s!$! --image-service-endpoint=/var/run/cri-resmgr/cri-resmgr.sock!' -i /etc/systemd/system/kubelet.service.d/10-kubeadm.conf `
```
justadogistaken, line 42: I guess replace cri-resmgr with sealer-cri-shim could be better. We are not using the complete cri-resmgr, are we?

shy-Xu: Yes, it's just a sock file name, we actually use sealer-cri-shim inside.

shy-Xu:
The CloudImage `kubernetes-crishim:v1.19.8`  changes images without specified registry to `sea.hub:5000/<image-path>:<image-tag>`, images with specified registry will not be processed

fanux:
I think the cri socket need not change any sealer code. Start the cri socket daemon in init.sh

PR #1150, opened 2022-02-17
feat:build support save output as tar file

kakaZhou719:

build default output is filesystem, if user specify "type=local,dest=dir or filename" ,will save output as local file.

save output at filesystem:

`sealer build -t my-kubernetes:1.19.8`

save output as local file:

`sealer build -t my-kubernetes:1.19.8 --output type=local,dest=/path/to/my-image.tar`

at meantime , `sealer save` support specify dir with default image name as tar filename.

`sealer save kubernetes:v1.19.8 -o /tmp `

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -15,6 +15,11 @@
 package build
 
 import (
+	"encoding/csv"
+	"fmt"
+	"strings"
+
+	"github.com/alibaba/sealer/build/buildkit/buildstorage"
 	"github.com/alibaba/sealer/build/cloud"
 	"github.com/alibaba/sealer/build/lite"
 	"github.com/alibaba/sealer/build/local"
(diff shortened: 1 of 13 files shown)


fanux:
We already has `sealer save` why need this feature?  Or why not just call the `Save` interface when complete build?

kakaZhou719:
> We already has `sealer save` why need this feature? Or why not just call the `Save` interface when complete build?

1, use "--output", we can support to set image local save dir , or save image tar file to oss or docker hub . and the default is local filesystem. 
2, old save interface  mix transfer imagename to *v1.image ,loading image metadata, writing file together, cannot be used directly.
3,`sealer save` dont support specify dir must the full ${image name} .tar .

PR #1169, opened 2022-02-22
Add autobuild script

Stevent-fei:

### DIFF
--- a/hack/autobuild.sh
+++ b/hack/autobuild.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+# Copyright © 2021 Alibaba Group Holding Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# -----------------------------------------------------------------------------
+#The script format is similar to that of auto build：
+#If we need to make kubernetes:1.20.14 image，We just need to enter: kubernetes v1.20.14 amd64
+
+echo "please enter image name:"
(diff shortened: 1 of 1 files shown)


hack/autobuild.sh
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# -----------------------------------------------------------------------------
+#The script format is similar to that of auto build：
+#If we need to make kubernetes:1.20.14 image，We just need to enter: kubernetes v1.20.14 amd64
+
+echo "please enter image name:"
+read  imageName   #eg: kubernetes v1.20.14 amd64
```
fanux, line 21: Read from stdin is not good. Just using `autobuild.sh kubernetes v1.20.14 amd64`  then using $1 $2 $3 to get args in your script

hack/autobuild.sh
```diff
@@ -0,0 +1,42 @@
...
+sudo cp ./kubernetes/server/bin/kubeadm ./rootfs/bin/
+sudo cp ./kubernetes/server/bin/kubelet ./rootfs/bin/
+wget https://dl.k8s.io/$version/kubernetes-server-linux-amd64.tar.gz
+tar -xvf kubernetes-server-linux-amd64.tar.gz
+wget http://sealer.oss-cn-beijing.aliyuncs.com/auto-build/sealer.tar.gz
+sudo tar -xvf sealer.tar.gz -C /usr/bin
+sudo sed -i "s/v1.20.14/$version/g" ./rootfs/etc/kubeadm.yml
+sudo sed -i "s/v1.20.14/$version/g" ./rootfs/Metadata
+sudo sed -i "s/amd64/$arch/g" ./rootfs/Metadata
+sudo ./kubernetes/server/bin/kubeadm config images list --config ./rootfs/etc/kubeadm.yml 2>/dev/null>>./rootfs/imageList
```
kakaZhou719, line 40: use the default imagerepo "k8s.gcr.io" of kubeadm.yml , will failed to upgrade from v1.19.8 to the latest. because  imagerepo of v1.19.8 is "sea.hub:5000/library" already saved at cluster configmap. this value will be loaded at the upgrade process. @fanux, We need to make sure whether we want to tag the image or modify the v1.19.8.

fanux (approved):

PR #1177, opened 2022-02-24
nydus: get localIP and set nydusd host config Automatically

nan1994:
Signed-off-by: lihongnan <hongnan.lhn@alibaba-inc.com>

Get localIP and set nydusd host config Automatically.
Nydusd need to set the IP of nydusdserver as host config in httpserver.json, and it is unfriendly to modify the nydusd configs manually.

NONE

### DIFF
--- a/applications/nydus/nydusdfile/serverstart.sh
+++ b/applications/nydus/nydusdfile/serverstart.sh
@@ -3,6 +3,36 @@
 #nydusdir need be scp
 rm -rf $2
 mkdir $2
+nydusdconfig='
+{\n
+  "device": {\n
+    "backend": {\n
+      "type": "registry",\n
+      "config": {\n
+        "scheme": "http",\n
+        "host": "'${3}':8000",\n
+        "repo": "sealer"\n
+      }\n
+    },\n
+    "cache": {\n
+      "type": "blobcache",\n
+      "config": {\n
+        "work_dir": "./cache"\n
+      }\n
+    }\n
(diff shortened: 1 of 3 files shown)


applications/nydus/nydusdfile/serverstart.sh
```diff
@@ -3,6 +3,36 @@
 #nydusdir need be scp
 rm -rf $2
 mkdir $2
+nydusdconfig='
+{\n
+  "device": {\n
+    "backend": {\n
+      "type": "registry",\n
```
kakaZhou719, line 10: here i think we can set  fixed "127.0.0.1" in  httpserver.json. and use `sed -i "s/127.0.0.1/$3/g" ./nydusd_scp_file/httpserver.json` to modify lcoal ip.

utils/iputils.go
```diff
@@ -79,6 +79,15 @@ func IsLocalIP(ip string, addrs *[]net.Addr) bool {
 	return false
 }
 
+func GetLocalIP(addrs *[]net.Addr) net.IP {
+	for _, address := range *addrs {
+		if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() && ipnet.IP.To4() != nil {
```
kakaZhou719, line 84: what if the host have mutil interface, we do better compare the ip net cidr with the master0ip to select the right one.

fanux (approved):

PR #1213, opened 2022-03-04
add clusterfile pre-processor

bxy4543:
add clusterfile pre-processor

### DIFF
--- a/pkg/clusterfile/pre_process.go
+++ b/pkg/clusterfile/pre_process.go
@@ -1,0 +1,176 @@
+// Copyright © 2021 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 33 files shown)


pkg/clusterfile/pre_process.go
```diff
@@ -0,0 +1,154 @@
...
+	c.Configs = configs
+	return nil
+}
+
+func (c *ClusterFile) PrePareCluster() error {
+	data, err := ioutil.ReadFile(c.path)
+	if err != nil {
+		return err
+	}
+	for _, d := range bytes.Split(data, []byte("---")) {
```
fanux, line 100: If user config contains "---" this may be had risk.

fanux: https://github.com/alibaba/sealer/blob/main/utils/file.go#L505

fanux (approved):

PR #1220, opened 2022-03-07
nydus: fix an error about async in mountNydusRootfs

nan1994:
When the node running sealer is one of masters, the node maybe run
nydusdInitCmd before scp nydud dir to other nodes finish. It will
scp all files in rootfs because the nydusd src dir is the same as
target dir. So change the nydusd src dir to be different from nydusdDir.

NONE


fanux:
PR to main branch

PR #1221, opened 2022-03-07
nydus:fix an error about async in mountNydusRootfs

nan1994:
bugfix: fix an error about async in mountNydusRootfs
When the node running sealer is one of masters, the node maybe run nydusdInitCmd before scp nydud dir to other nodes finish. It will scp all files in rootfs because the nydusd src dir is the same as target dir. So change the nydusd src dir to be different from nydusdDir.

NONE


pkg/filesystem/cloudfilesystem/nydus.go
```diff
@@ -120,7 +121,7 @@ func mountNydusRootfs(ipList []string, target string, cluster *v2.Cluster, initF
 			if err != nil {
 				return fmt.Errorf("get host ssh client failed %v", err)
 			}
-			err = copyFiles(sshClient, ip == config.IP, ip, nydusdDir, nydusdDir)
```
fanux, line 123: If CloudImage is not a nydus CloudImage, will send registry dir to all nodes.

fanux (approved):

PR #1335, opened 2022-04-25
add gpu operator on containerd cloudImage

fanux:


applications/gpu-operator/Kubefile
```diff
@@ -0,0 +1,8 @@
+FROM scratch
+COPY . .
+RUN helm repo add nvidia https://helm.ngc.nvidia.com/nvidia \
```
kakaZhou719, line 3: need copy helm binary file.

applications/gpu-operator/Kubefile
```diff
@@ -0,0 +1,8 @@
+FROM scratch
+COPY . .
```
kakaZhou719, line 2: COPY imageList manifests

PR #1342, opened 2022-04-27
[wip]feat:no need to set ip and password for sealer run

kakaZhou719:
### Describe how you did it

For example, I have logged in to a machine which as my master0 node, and my cluster only have this one node.

and then I just want to initialize the machine. At this time, I don’t need to enter the password and ip addresses . cmd "sealer run " will success .

Example : "sealer run kubernetes:v1.19.8"


fanux (approved):

fanux:
awesome feature~

Issue #1406, opened 2022-05-16
the registry certificate of the non master node was not updated when sealer gen


sfeng1996:
#### What happened:

./sealer gen --image kubernetes:v1.19.8 --passwd xxxxxx --name test

only master0 has updated the registry certificate，Error in docker pull or docker push  of other nodes

error :
```
$ docker pull sea.hub:5000/tigera/operator:v1.25.3
Error response from daemon: Get https://sea.hub:5000/v2/: x509: certificate signed by unknown authority

```

#### Environment:

- sealer version (use `sealer version`):
v0.8.5

PR #1408, opened 2022-05-16
fix: the registry certificate of the non master node was not updated …

sfeng1996:
https://github.com/sealerio/sealer/issues/1406

### DIFF
--- a/apply/processor/gen.go
+++ b/apply/processor/gen.go
@@ -187,7 +187,15 @@ func (g *GenerateProcessor) ApplyRegistry(cluster *v2.Cluster) error {
 	if err != nil {
 		return err
 	}
-	err = runt.(*runtime.KubeadmRuntime).GenerateRegistryCert()
+	rt, ok := runt.(*runtime.KubeadmRuntime)
+	if !ok {
+		return fmt.Errorf("invalid type")
+	}
+	err = rt.GenerateRegistryCert()
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


bxy4543 (approved):

apply/processor/gen.go
```diff
@@ -187,7 +187,16 @@ func (g *GenerateProcessor) ApplyRegistry(cluster *v2.Cluster) error {
...
-	err = runt.(*runtime.KubeadmRuntime).GenerateRegistryCert()
+	rt, ok := runt.(*runtime.KubeadmRuntime)
+	if !ok {
+		return fmt.Errorf("valid runtime")
+	}
+	err = rt.GenerateRegistryCert()
+	if err != nil {
+		return err
+	}
+	// sealer need to update registry cert on other nodes, otherwise docker pull or push will report error
```
kakaZhou719, line 198: here i think ,could we send to all nodes whether it is master0 or not after generate registry cert

fanux: agress with you

sfeng1996: ok，I'd like to ask how to realize it
send to master0  has been implemented in rt.GenerateRegistryCert()

kakaZhou719: don't worry about this，how about we split it ? method "GenerateRegistryCert" only generate registry cert, and we use another method to do send .

PR #1409, opened 2022-05-16
optimize: parse the custom resource definitions method refactoring

bxy4543:
Resolve custom resource definition method is too bloated, optimize the structure, support different versions of the resolution of different structures.
kubeadm version v1beta3 structure support later.

Signed-off-by: wb-hjh933779 <wb-hjh933779@alibaba-inc.com>


utils/yaml/decode.go
```diff
@@ -0,0 +1,224 @@
...
+	"k8s.io/kubelet/config/v1beta1"
+
+	"github.com/sealerio/sealer/common"
+	"github.com/sealerio/sealer/logger"
+	v1 "github.com/sealerio/sealer/types/api/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+var decodeCRDFuncMap = map[string]func(reader io.Reader) (interface{}, error){
```
kakaZhou719, line 39: those logic to do decode file  not match with the pkg name `yaml` means, i think we can move it to decode pkg or another.

kakaZhou719 (approved):

PR #1455, opened 2022-06-01
feat: add more checker for cluster init

kakaZhou719:


pkg/checker/os_checker.go
```diff
@@ -0,0 +1,87 @@
...
+			if r.OSName == os.OSName &&
+				strings.HasPrefix(os.OSVersion, r.OSVersion) &&
+				strings.HasPrefix(os.KernelVersion, r.KernelVersion) &&
+				r.Arch == os.Arch {
+				match = true
+				break
+			}
+		}
+		if !match {
+			return fmt.Errorf("the current host is: \n%s.\nthe OS only support: \n%s", os.GetStr(), strings.Join(GetSupportedOSStr(), ",\n"))
```
fanux, line 83: I think should not return error, just warning, because sealer already support other linux distribution.

justadogistaken:
I'm considering if we could make some checkers be plugins. Design some interfaces, write them in golang instead of shell(I think golang is capable of complete what shell can do). Make the plugins impletation like scheduler framework.

PR #1465, opened 2022-06-06
fix Clusterfile not create bug

Stevent-fei:

### DIFF
--- a/apply/processor/gen.go
+++ b/apply/processor/gen.go
@@ -18,7 +18,6 @@ package processor
 
 import (
 	"fmt"
-	"os"
 	"strconv"
 
 	"github.com/sealerio/sealer/utils/yaml"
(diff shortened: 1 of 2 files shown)


utils/yaml/yaml.go
```diff
@@ -74,6 +74,11 @@ func MarshalToFile(file string, obj interface{}) error {
 		return err
 	}
 
+	create, _ := os.Create("Clusterfile")
```
bxy4543, line 77: ?

apply/processor/gen.go
```diff
@@ -74,6 +75,12 @@ func NewGenerateProcessor() (Processor, error) {
 
 func (g *GenerateProcessor) init(cluster *v2.Cluster) error {
 	fileName := fmt.Sprintf("%s/.sealer/%s/Clusterfile", common.GetHomeDir(), cluster.Name)
+	clusterPath := fmt.Sprintf("%s/.sealer/%s", common.GetHomeDir(), cluster.Name)
+	err := os.MkdirAll(clusterPath, 0750)
```
bxy4543, line 79: add this in the yaml.MarshalToFile function.

utils/yaml/yaml.go
```diff
@@ -74,6 +74,11 @@ func MarshalToFile(file string, obj interface{}) error {
...
+	defer func() {
+		_ = create.Close()
+	}()
+
 	if err = os.WriteFile(file, data, common.FileMode0644); err != nil {
```
bxy4543, line 82: Just use the WriteFile method of the utils/os/writers.go path

bxy4543 (approved):

PR #1472, opened 2022-06-07
Remove useless config Level in logger;

VinceCui:
Support remote logger, which support sealer to continuosly post logs to a remote URL.


VinceCui:
I will use logrus hook to implement this feature.

PR #1588, opened 2022-07-15
upgrade IsIpv4 method

zsyaoo:
use golang net package api to upgrade IsIpv4 method.

### DIFF
--- a/utils/net/route.go
+++ b/utils/net/route.go
@@ -158,28 +158,5 @@ func delRouteGatewayViaHost(host, gateway net.IP) error {
 }
 
 func IsIpv4(ip string) bool {
-	arr := strings.Split(ip, ".")
-	if len(arr) != 4 {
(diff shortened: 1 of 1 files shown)


justadogistaken:
@zsyaoo Great! But do we have some tests for this function?

zsyaoo:
> @zsyaoo Great! But do we have some tests for this function?

I think `system api` or `simple code` no more need to tests

zsyaoo:
> @zsyaoo Ok, I checked there is iputils_test.go. Actually I'm not talking about testing `net.ParseIP`, it's about testing the upper function. And there it is. Thanks.

IsIpv4 function are not used , i think it will be delete

zsyaoo:
IsIpv4 function are not used,we can use `k8s.io/utils/net` IsIPv4String() function

PR #1603, opened 2022-07-26
[WIP]modify unit test

Stevent-fei:
Modify some obsolete methods in unit tests

### DIFF
--- a/pkg/env/env_test.go
+++ b/pkg/env/env_test.go
@@ -64,6 +64,23 @@ func getTestCluster() *v2.Cluster {
 	}
 }
 
+func getTestClusterWrapperShell() *v2.Cluster {
+	return &v2.Cluster{
+		Spec: v2.ClusterSpec{
+			Image: "",
+			Env:   []string{"key=value"},
+			Hosts: []v2.Host{
+				{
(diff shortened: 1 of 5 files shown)


pkg/env/env_test.go
```diff
@@ -64,7 +64,24 @@ func getTestCluster() *v2.Cluster {
 	}
 }
 
-/* func Test_processor_WrapperShell(t *testing.T) {
+func getTestClusterWrapperShell() *v2.Cluster {
```
kakaZhou719, line 67: getClusterSpecWithEnv? feel free to describe the function name as clearly as possible.

pkg/plugin/plugins_test.go
```diff
@@ -14,27 +14,25 @@
...
+
+	v1 "github.com/sealerio/sealer/types/api/v1"
+	v2 "github.com/sealerio/sealer/types/api/v2"
+)
+
 func TestDumperPlugin_Run(t *testing.T) {
 	type fields struct {
-		configs     []v1.Plugin
-		clusterName string
+		Plugins []v1.Plugin
```
kakaZhou719, line 27: lower case the  struct fields if only used internally。`plugins`

PR #1622, opened 2022-08-08
add ssh infra driver

kakaZhou719:

Through the re-encapsulation of the ssh package, the cluster operation can be completed directly through the ip address, instead of passing the v2.cluster structure to the ssh inner layer as before we do.

### DIFF
--- a/pkg/infradriver/ssh_infradriver.go
+++ b/pkg/infradriver/ssh_infradriver.go
@@ -50,7 +50,12 @@ func NewInfraDriver(cluster *v2.Cluster) (InfraDriver, error) {
 	// initialize roleHostsMap field
 	for _, host := range cluster.Spec.Hosts {
 		for _, role := range host.Roles {
-			ret.roleHostsMap[role] = host.IPS
+			ips, ok := ret.roleHostsMap[role]
+			if !ok {
+				ret.roleHostsMap[role] = host.IPS
+			} else {
+				ret.roleHostsMap[role] = append(ips, host.IPS...)
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/infradriver/ssh_infradriver.go
```diff
@@ -1,56 +1,143 @@
...
+		}
+		for _, ip := range host.IPS {
+			ret.sshConfigs[ip.String()] = ssh.NewSSHClient(&host.SSH, true)
+		}
+	}
+
+	// initialize roleHostsMap field
+	for _, host := range cluster.Spec.Hosts {
+		for _, role := range host.Roles {
+			ret.roleHostsMap[role] = host.IPS
```
VinceCui, line 53: need append, because different host maybe have same role.

VinceCui: Any update？

kakaZhou719: it already fixed, pls review.

pkg/infradriver/ssh_infradriver.go
```diff
@@ -1,56 +1,167 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 package infradriver
 
 import (
+	"fmt"
+	"github.com/imdario/mergo"
```
VinceCui, line 19: group them

VinceCui (approved):
LGTM

PR #1629, opened 2022-08-11
[WIP]Add container runtime driver

kakaZhou719:

Add container runtime driver:  support to config container runtime, like access private registry.


pkg/container-runtime/driver.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (d DockerRuntimeDriver) configRegistryCert(domain, port string, caFile string, ips []net.IP) error {
+	// copy ca cert to "/etc/docker/certs.d/${domain}:${port}/
+}
+
+func (d DockerRuntimeDriver) configDaemon(endpoint string, ips []net.IP) error {
+	// modify daemon.json: "mirrors": ["${endpoint}"],docker server need to restart if modify daemon.json
+}
+
+func NewDockerRuntimeDriver(ssh infradriver.InfraDriver) ContainerRuntimeDriver {
```
VinceCui, line 72: Don't name as "ssh"

pkg/container-runtime/driver.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+//ContainerRuntimeDriver provide configuration Interface for different container runtime.
+type ContainerRuntimeDriver interface {
+	// ConfigRegistry config registry via each container runtime ip address
+	ConfigRegistry(ConfigForRegistry, []net.IP) error
+}
+
+type DockerRuntimeDriver struct {
+	ssh infradriver.InfraDriver
```
VinceCui, line 45: Too

pkg/container-runtime/driver.go
```diff
@@ -0,0 +1,76 @@
...
+// limitations under the License.
+
+package container_runtime
+
+import (
+	"github.com/sealerio/sealer/pkg/infradriver"
+	"net"
+)
+
+type ConfigForRegistry struct {
```
VinceCui, line 22: Why should ContainerRuntimeInstaller know ConfigForRegistry?

kakaZhou719: this is not for `ContainerRuntimeInstaller`, its `Configurator` for ContainerRuntime, for example: config ContainerRuntime to access local registry.

PR #1652, opened 2022-08-18
Add Container runtime---docker install

Stevent-fei:
Install docker:
Modify the docker configuration file through shell commands
Execute the clean script file when uninstalling

### DIFF
--- a/pkg/container-runtime/dockerInstaller.go
+++ b/pkg/container-runtime/dockerInstaller.go
@@ -44,7 +44,7 @@ func (d *DockerInstaller) InstallOn(hosts []net.IP) (*Info, error) {
 		initCmd := fmt.Sprintf(RemoteChmod, d.rootfs, DefaultDomain, DefaultPort, d.Info.CgroupDriver, d.Info.LimitNofile)
 		err := d.driver.CmdAsync(ip, initCmd)
 		if err != nil {
-			return nil, fmt.Errorf("failed to remote exec init cmd: %s", err)
+			return nil, fmt.Errorf("failed to exec the install docker init command remote: %s", err)
 		}
 	}
 	return info, nil
@@ -55,7 +55,7 @@ func (d *DockerInstaller) UnInstallFrom(hosts []net.IP) error {
 	for _, ip := range hosts {
 		err := d.driver.CmdAsync(ip, CleanCmd)
 		if err != nil {
-			return fmt.Errorf("failed to remote exec clean cmd: %s", err)
+			return fmt.Errorf("failed to exec clean docker command remote: %s", err)
 		}
 	}
 	return nil
--- a/pkg/container-runtime/dockerInstaller.go
+++ b/pkg/container-runtime/dockerInstaller.go
@@ -28,15 +28,17 @@ type DockerInstaller struct {
 }
 
 func (d *DockerInstaller) InstallOn(hosts []net.IP) (*Info, error) {
-	RemoteChmod := "cd %s  && chmod +x scripts/* && cd scripts && bash init.sh /var/lib/docker %s %s %s %s"
+	RemoteChmod := "cd %s  && chmod +x scripts/* && cd scripts && bash init.sh /var/lib/docker %s %s"
 	info := &Info{
-		Config: Config{
-			LimitNofile:  DefaultLimitNoFile,
-			CgroupDriver: DefaultSystemdDriver,
+		Config{
+			Docker,
+			DefaultLimitNoFile,
+			DefaultSystemdDriver,
 		},
+		DefaultDockerSocket,
 	}
 	for _, ip := range hosts {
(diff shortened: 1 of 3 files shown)


pkg/container-runtime/dockerInstaller.go
```diff
@@ -0,0 +1,70 @@
...
+package container_runtime
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/sealerio/sealer/pkg/infradriver"
+)
+
+type ContainerdInstaller struct {
```
allencloud, line 24: If this file is named by dockerInstaller.go, then ContainerdInstaller should be located in another file which is named by containerdInstaller.go.

In addition, the file structure should be :
pkg/container-runtime/installer/docker.go 
pkg/container-runtime/installer/containerd.go

pkg/container-runtime/installer.go
```diff
@@ -35,8 +50,21 @@ type Info struct {
...
-//type containerdInstaller struct{}
-type customContainerRuntimeInstaller struct{}
+	if conf.Type == Containerd {
+		containerdInstaller := &ContainerdInstaller{
+			rootfs: driver.GetClusterRootfs(),
+			driver: driver,
+		}
+		return containerdInstaller
+	}
+	return nil
```
allencloud, line 69: this line will cause severe panic. This function cannot return a nil. This function should be defined to return two parameter
```
func NewInstaller(conf Config, driver infradriver.InfraDriver) (Installer, error) {
```
When the input config type is neither docker nor containerd. This function should return an error.

pkg/container-runtime/dockerInstaller.go
```diff
@@ -0,0 +1,57 @@
...
+type DockerInstaller struct {
+	Info   Info
+	rootfs string
+	driver infradriver.InfraDriver
+}
+
+func (d *DockerInstaller) InstallOn(hosts []net.IP) (*Info, error) {
+	RemoteChmod := "cd %s  && chmod +x scripts/* && cd scripts && bash init.sh /var/lib/docker %s %s %s %s"
+	info := &Info{
+		Config: Config{
```
kakaZhou719, line 33: can be simplified

pkg/container-runtime/dockerInstaller.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type DockerInstaller struct {
+	Info   Info
+	rootfs string
+	driver infradriver.InfraDriver
+}
+
+func (d *DockerInstaller) InstallOn(hosts []net.IP) (*Info, error) {
+	RemoteChmod := "cd %s  && chmod +x scripts/* && cd scripts && bash init.sh /var/lib/docker %s %s %s %s"
```
kakaZhou719, line 31: In my opinion, the parameters of the init.sh script only need to be related to the container runtime installation process. so, `/var/lib/docker`,` DefaultDomain`, `DefaultPort`, Are these parameters unnecessary?

pkg/container-runtime/dockerInstaller.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+type DockerInstaller struct {
+	Info   Info
+	rootfs string
+	driver infradriver.InfraDriver
+}
+
+func (d *DockerInstaller) InstallOn(hosts []net.IP) (*Info, error) {
+	RemoteChmod := "cd %s  && chmod +x scripts/* && cd scripts && bash init.sh /var/lib/docker %s %s"
```
VinceCui, line 31: why init.sh?

pkg/container-runtime/dockerInstaller.go
```diff
@@ -0,0 +1,59 @@
...
+		err := d.driver.CmdAsync(ip, initCmd)
+		if err != nil {
+			return nil, fmt.Errorf("failed to exec the install docker init command remote: %s", err)
+		}
+	}
+	return info, nil
+}
+
+func (d *DockerInstaller) UnInstallFrom(hosts []net.IP) error {
+	CleanCmd := "cd %s  && chmod +x scripts/* && cd scripts && bash clean.sh"
```
VinceCui, line 51: why clean.sh?

pkg/container-runtime/dockerInstaller.go
Stevent-fei: init.sh is the script entry for installing docker and k8s, which supports passing parameters to change some configuration files
https://github.com/sealerio/basefs/blob/main/context/rootfs/scripts/init.sh

pkg/container-runtime/dockerInstaller.go
Stevent-fei: clean.sh is the script to uninstall docker/containerd

pkg/container-runtime/dockerInstaller.go
VinceCui: just use docker.sh

pkg/container-runtime/dockerInstaller.go
VinceCui: image

should use docker-uninstall.sh

PR #1674, opened 2022-08-29
[optimization] make decoding clusterfile more simple and readable

kakaZhou719:

1. reconstruct clusterfile , it is unnecessary to abstract `PreProcessor`.  i  just parse and construct clusterfile object.
2.  load each CRD object from `clusterfile` with "---" as its end line.
3.  for plugin and config resource, Load it  in the form of a list of resources.


pkg/clusterfile/clusterfile.go
```diff
@@ -15,61 +15,59 @@
...
 	}
-	once.Do(func() {
-		clusterFile.path = path
-		err = clusterFile.Process()
-	})
-	return clusterFile, err
+
+	clusterFile := new(ClusterFile)
+	err = decodeClusterFile(bytes.NewReader(clusterFileData), clusterFile)
+
```
justadogistaken, line 67: remove the space.

justadogistaken (approved):

PR #1675, opened 2022-08-30
add run and join

kakaZhou719:
1.  add run cmd and join cmd 
2. decompose filesystem into a distributor
3. optimization utils filesystem rename func and its test case.


pkg/infradriver/utils.go
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+
+package infradriver
+
+import (
+	v1 "github.com/sealerio/sealer/types/api/v1"
+	"net"
+)
+
+func GetHostsPlatform(hosts []net.IP, driver InfraDriver) (map[v1.Platform][]net.IP, error) {
```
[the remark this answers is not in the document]
kakaZhou719: yes, this  function should belong to InfraDriver. i will update this later.

cmd/sealer/cmd/run.go
```diff
@@ -76,12 +89,66 @@ create a cluster with custom environment variables:
...
+			return err
+		}
+
+		var (
+			clusterImageName  = cluster.Spec.Image
+			rootfs            = infraDriver.GetClusterRootfs()
+			rootfsRegistryDir = filepath.Join(rootfs, "registry")
+		)
+
+		for platform, hosts := range hostsPlatformMap {
```
VinceCui, line 119: hostsPlatformMap should be handle by imageEngine and imageDistributor, not cmd run.

pkg/imagedistributor/nydus_distributor.go
```diff
@@ -0,0 +1,15 @@
+// Copyright © 2022 Alibaba Group Holding Ltd.
+//
```
VinceCui, line 2: now just skip nydus

pkg/infradriver/infradriver.go
```diff
@@ -26,15 +26,17 @@ type InfraDriver interface {
 
 	GetHostIPListByRole(role string) []net.IP
 
+	//GetHostPlatform by ip address
+	GetHostPlatform(host net.IP) (v1.Platform, error)
```
VinceCui, line 30: no need this interface, just use GetHostsPlatform([]net.IP{host})

cmd/sealer/cmd/run.go
```diff
@@ -76,12 +89,66 @@ create a cluster with custom environment variables:
...
+			if err := pullClusterImage(clusterImageName, platform, imageEngine); err != nil {
+				return err
+			}
+
+			// mount cluster image
+			if err := mountClusterImage(clusterImageName, mountDir, imageEngine); err != nil {
+				return err
+			}
+
+			// send registry to local host dir.
```
VinceCui, line 135: how to send registry dir, should be implemented by RegistryConfigurator, imageEngine only needs to be responsible for preparing the registry directory.

cmd/sealer/cmd/join.go
```diff
@@ -57,7 +59,27 @@ join default cluster:
...
+			return err
+		}
+
+		var (
+			clusterImageName = cluster.Spec.Image
+			hosts            = append(newMasters, newWorkers...)
+		)
+
+		// distribute rootfs
+		if err = distributor.Distribute(clusterImageName, imagedistributor.FilterOptions{
```
VinceCui, line 78: i think maybe we can rename func Distribute to DistributeRootfsDir.

in cmd, we just run DistributeRootfsDir, which only distribute rootfs except registry dir. And we give infradriver as a parameter of RegistryConfigurator, than RegistryConfigurator will use infradriver to copy registry dir or other things, depends on its own implementation

VinceCui: Because "			ExceptDirs: []string{"registry"}," is very confused.

cmd/sealer/cmd/join.go
```diff
@@ -57,12 +59,37 @@ join default cluster:
...
+			hosts            = append(newMasters, newWorkers...)
+		)
+
+		// distribute rootfs
+		if err = distributor.Distribute(clusterImageName, hosts); err != nil {
+			return err
+		}
+
+		runtimeConfig := new(clusterruntime.RuntimeConfig)
+		if cf.GetPlugins() != nil {
```
VinceCui, line 83: First of all, I think it is not allowed to re specify a new clusterfile when joining. You must directly use the clusterfile used in init cluster, so that you can ensure that there is only one global clusterfile, which is simpler.

Therefore, the clusterfile object here should directly obtain the existing clusterfile in the cluster (a copy of the clusterfile should be kept in a k8s configmap after the cluster init is completed)

pkg/registry/singleton.go
```diff
@@ -52,54 +58,102 @@ func (c *localSingletonConfigurator) Clean() error {
 	return c.infraDriver.CmdAsync(c.DeployHost, fmt.Sprintf(deleteRegistryCommand, "sealer-registry"))
 }
 
+func (c *localSingletonConfigurator) UninstallFrom(hosts []net.IP) error {
+	cmd := fmt.Sprintf("rm -f /root/.docker/config.json && %s", shellcommand.CommandUnSetHostAlias())
```
VinceCui, line 62: why can RegistryConfigurator remove docker cli config?

if you want to clean the login credentials of Registry from docker, use docker logout.

cmd/sealer/cmd/join.go
kakaZhou719: What I think here is that the clusterfile in the `sealer run` phase is stored locally, and then when execute `sealer join`, hosts args will be merged with the local file. That is to say, there is no need to specify the clusterfile file in the `sealer join` and `sealer delete` cmds.  what do you think ?

PR #1685, opened 2022-09-04
fix: set a default storage conf

justadogistaken:
Screen Shot 2022-09-04 at 20 35 30

Set a default storage config for containers/images

### DIFF
--- a/pkg/imageengine/buildah/init.go
+++ b/pkg/imageengine/buildah/init.go
@@ -56,6 +56,41 @@ registries = []
 	"auths": {}
 }
 `
+
+	buildahStorageConf = `
+# storage.conf is the configuration file for all tools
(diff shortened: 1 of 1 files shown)


pkg/imageengine/buildah/init.go
```diff
@@ -86,6 +121,11 @@ func initBuildah() error {
 		return err
 	}
 
+	storageAbsPath := "/etc/containers/storage.conf"
```
kakaZhou719, line 124: do we need to mark storageAbsPath as a const ?

justadogistaken: It could be the const. But currently I think this will be used in one place only. I guess make it a temporary var is acceptable.

kakaZhou719 (approved):

PR #1686, opened 2022-09-05
Feature: k0s runtime remain work

starComingup:
After k0s init step, there are update/join/delete/reset interface which need to implement, this PR finish all k0s cluster install action.

NONE, but a step for #1399

### DIFF
--- a/apply/processor/utils.go
+++ b/apply/processor/utils.go
@@ -1,0 +1,39 @@
+// Copyright © 2022 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package processor
+
+import (
+	"github.com/sealerio/sealer/pkg/runtime"
+	"github.com/sealerio/sealer/pkg/runtime/k0s"
+	"github.com/sealerio/sealer/pkg/runtime/kubernetes"
+	"github.com/sealerio/sealer/pkg/runtime/kubernetes/kubeadm"
+	v2 "github.com/sealerio/sealer/types/api/v2"
+)
+
+func RuntimeChoose(rootfs string, cluster *v2.Cluster, config *kubeadm.KubeadmConfig) (runtime.Interface, error) {
+	metadata, err := runtime.LoadMetadata(rootfs)
+	if err != nil {
+		return nil, err
+	}
+	switch metadata.ClusterRuntime {
+	case runtime.K8s:
+		return kubernetes.NewDefaultRuntime(cluster, config)
+	case runtime.K0s:
+		return k0s.NewK0sRuntime(cluster)
+	// Todo case runtime.K3s:
+	default:
+		return kubernetes.NewDefaultRuntime(cluster, config)
+	}
+}
--- a/apply/processor/utils.go
+++ b/apply/processor/utils.go
@@ -20,13 +20,17 @@ import (
 	"github.com/sealerio/sealer/pkg/runtime/kubernetes"
 	"github.com/sealerio/sealer/pkg/runtime/kubernetes/kubeadm"
 	v2 "github.com/sealerio/sealer/types/api/v2"
+	"github.com/sirupsen/logrus"
 )
 
-func RuntimeChoose(rootfs string, cluster *v2.Cluster, config *kubeadm.KubeadmConfig) (runtime.Interface, error) {
+func ChooseRuntime(rootfs string, cluster *v2.Cluster, config *kubeadm.KubeadmConfig) (runtime.Interface, error) {
 	metadata, err := runtime.LoadMetadata(rootfs)
 	if err != nil {
 		return nil, err
 	}
+
+	hintInfo(metadata.ClusterRuntime)
+
 	switch metadata.ClusterRuntime {
 	case runtime.K8s:
 		return kubernetes.NewDefaultRuntime(cluster, config)
@@ -36,4 +40,14 @@ func RuntimeChoose(rootfs string, cluster *v2.Cluster, config *kubeadm.KubeadmCo
 	default:
 		return kubernetes.NewDefaultRuntime(cluster, config)
 	}
+}
+
+// hintInfo will hint end-user which cluster runtime will load and move to using new version ClusterImage.
+func hintInfo(clusterRuntime runtime.ClusterRuntime) {
+	if clusterRuntime != "" {
+		logrus.Infof("using %s as a cluster runtime.", clusterRuntime)
+		return
+	}
+	logrus.Infof("using k8s as a default cluster runtime.")
+	logrus.Infof("the old sealer version is deprecated, using new version of sealer and ClusterImage is encouraged!")
 }
--- a/pkg/runtime/k0s/delete_masters.go
+++ b/pkg/runtime/k0s/delete_masters.go
@@ -1,0 +1,124 @@
+// Copyright © 2022 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k0s
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	RemoteCleanMasterOrNode = `if which k0s; then k0s stop && k0s reset --config %s --cri-socket %s;fi && \
+rm -rf /etc/k0s/
+`
+	RemoveKubeConfig          = "rm -rf /usr/bin/kube* && rm -rf ~/.kube/"
+	RemoveK0sBin              = "rm -rf /usr/bin/k0s"
+	RemoteRemoveEtcHost       = "sed -i \"/%s/d\" /etc/hosts"
+	RemoteRemoveRegistryCerts = "rm -rf " + DockerCertDir + "/%s*"
+	KubeDeleteNode            = "kubectl delete node %s"
+)
+
+func (k *Runtime) deleteMasters(masters []net.IP) error {
+	if len(masters) == 0 {
+		return nil
+	}
+	eg, _ := errgroup.WithContext(context.Background())
+	for _, master := range masters {
+		master := master
+		eg.Go(func() error {
+			master := master
+			logrus.Infof("Start to delete master %s", master)
+			if err := k.deleteMaster(master); err != nil {
+				logrus.Errorf("failed to delete master %s: %v", master, err)
+			} else {
+				logrus.Infof("Succeeded in deleting master %s", master)
+			}
+			return nil
+		})
+	}
+	return eg.Wait()
+}
+
+func (k *Runtime) deleteMaster(master net.IP) error {
+	ssh, err := k.getHostSSHClient(master)
+	if err != nil {
+		return fmt.Errorf("failed to delete master: %v", err)
+	}
+	remoteCleanCmd := []string{fmt.Sprintf(RemoteCleanMasterOrNode, DefaultK0sConfigPath, ExternalCRI),
+		fmt.Sprintf(RemoteRemoveEtcHost, SeaHub),
+		fmt.Sprintf(RemoteRemoveRegistryCerts, k.RegConfig.Domain),
+		fmt.Sprintf(RemoteRemoveRegistryCerts, SeaHub),
+		RemoveKubeConfig,
+		RemoveK0sBin}
+
+	if err := ssh.CmdAsync(master, remoteCleanCmd...); err != nil {
+		return err
+	}
+
+	// remove master
+	masterIPs := []net.IP{}
+	for _, ip := range k.cluster.GetMasterIPList() {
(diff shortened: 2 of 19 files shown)


pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,124 @@
...
+
+const (
+	RemoteCleanMasterOrNode = `if which k0s; then k0s stop && k0s reset --config %s --cri-socket %s;fi && \
+rm -rf /etc/k0s/
+`
+	RemoveKubeConfig          = "rm -rf /usr/bin/kube* && rm -rf ~/.kube/"
+	RemoveK0sBin              = "rm -rf /usr/bin/k0s"
+	RemoteRemoveEtcHost       = "sed -i \"/%s/d\" /etc/hosts"
+	RemoteRemoveRegistryCerts = "rm -rf " + DockerCertDir + "/%s*"
+	KubeDeleteNode            = "kubectl delete node %s"
```
justadogistaken, line 35: I'd rather put those command at the caller place. This will break the readability.

pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (k *Runtime) deleteMasters(masters []net.IP) error {
+	if len(masters) == 0 {
+		return nil
+	}
+	eg, _ := errgroup.WithContext(context.Background())
+	for _, master := range masters {
+		master := master
+		eg.Go(func() error {
+			master := master
```
justadogistaken, line 46: I guess there is no need to declare the temp var again.

pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+const (
+	RemoteCleanMasterOrNode = `if which k0s; then k0s stop && k0s reset --config %s --cri-socket %s;fi && \
+rm -rf /etc/k0s/
+`
+	RemoveKubeConfig          = "rm -rf /usr/bin/kube* && rm -rf ~/.kube/"
+	RemoveK0sBin              = "rm -rf /usr/bin/k0s"
+	RemoteRemoveEtcHost       = "sed -i \"/%s/d\" /etc/hosts"
+	RemoteRemoveRegistryCerts = "rm -rf " + DockerCertDir + "/%s*"
```
allencloud, line 34: Please do not use format in the const string value. It would make code quite complicated.

pkg/runtime/k0s/delete_masters.go
allencloud, line 35: It should use an API calling torwards apiserver, rather than using a shell command.

pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil
+	}
+	eg, _ := errgroup.WithContext(context.Background())
+	for _, master := range masters {
+		master := master
+		eg.Go(func() error {
+			master := master
+			logrus.Infof("Start to delete master %s", master)
+			if err := k.deleteMaster(master); err != nil {
+				logrus.Errorf("failed to delete master %s: %v", master, err)
```
allencloud, line 49: You should `return err` here, otherwise the eg.Go() could not catch the detailed error, and the `return eg.Wait()` would always `return nil`, which is definitely unreasonable.

pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,124 @@
...
+			return nil
+		})
+	}
+	return eg.Wait()
+}
+
+func (k *Runtime) deleteMaster(master net.IP) error {
+	ssh, err := k.getHostSSHClient(master)
+	if err != nil {
+		return fmt.Errorf("failed to delete master: %v", err)
```
allencloud, line 62: I think your error log message should be optimized. When logging error message, you should set up a rule for yourself, where to locate the wrap message. In this case, the wrap message is `failed to delete master:`.  And in Line 49, you have code `logrus.Errorf("failed to delete master %s: %v", master, err)`. Then if an error is returned back, then you error message would be `failed to delete master 1.1.1.1: failed to delete master: xxxxxx`. It is duplicated and definitely inproper.

I wish that you could pay more attention on every part of the code. Only by patience and carefulness, could quality of software be achieved.

pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,124 @@
...
+			} else {
+				logrus.Infof("Succeeded in deleting master %s", master)
+			}
+			return nil
+		})
+	}
+	return eg.Wait()
+}
+
+func (k *Runtime) deleteMaster(master net.IP) error {
```
allencloud, line 59: This is a detailed deep implementation of delete a master node. I think we should explain clearly what is the **STEPS** to delete a master by comments. If we understand the steps and then the code review could be quite easy. Actually we should review the design firstly, and code secondly.  No design there, and we can hardly do the code review. Do you think so ?

pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,124 @@
...
+
+		if err := master0SSH.CmdAsync(k.cluster.GetMaster0IP(), fmt.Sprintf(KubeDeleteNode, strings.TrimSpace(hostname))); err != nil {
+			return fmt.Errorf("failed to delete node %s: %v", hostname, err)
+		}
+	}
+	return nil
+}
+
+func (k *Runtime) isHostName(master, host net.IP) (string, error) {
+	hostString, err := k.CmdToString(master, "kubectl get nodes | grep -v NAME  | awk '{print $1}'", ",")
```
allencloud, line 101: Please use k8s apiClient to get the information you want. Otherwise it is quite strange to add shell action in golang code. If KubeCtl released a higher version, and the output format changes, the code here will immediately meet the compatibility issue. What a bad design. Right?

allencloud:
I just reviewed the `delete_master.go` file, and attached several feedbacks. I think you have the same issue in other files, like `delete_node.go`, `join_node.go` and so on. Please update the original code all all files carefully. Then you could comment this pr to seek my further review. Thanks. @starComingup 🌹 🐯 🦁

pkg/runtime/k0s/delete_masters.go
starComingup: Yep! when I read kubernetes runtime module, I have the same feel. which files to delete? and what logic to follow. I so sorry to made this, fix this ASAP.

pkg/runtime/k0s/delete_masters.go
starComingup: Got it! next commit will fix it!

pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	/** To delete a node from k0s cluster, following these steps.
+	STEP1: stop k0s service
+	STEP2: reset the node with install configuration
+	STEP3: remove k0s cluster config generate by k0s under /etc/k0s
+	STEP4: remove private registry config in /etc/host
+	STEP5: remove bin file such as: kubectl, and remove .kube directory
+	STEP6: remove k0s bin file
+	STEP7: delete node though k8s client
+	*/
```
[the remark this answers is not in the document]
starComingup: I think the error will be thrown by ssh.Async and captured by the caller.

pkg/runtime/k0s/delete_masters.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if err := ssh.CmdAsync(master, remoteCleanCmd...); err != nil {
+		return err
+	}
+
+	// remove master
+	masterIPs := []net.IP{}
+	for _, ip := range k.cluster.GetMasterIPList() {
+		if !ip.Equal(master) {
+			masterIPs = append(masterIPs, ip)
```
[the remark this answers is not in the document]
starComingup: I read this code block and I think It did a thing that judge the rest master nodes need greater than one, you code is more clearly and simplify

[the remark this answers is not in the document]
starComingup: As for `len(k.cluster.GetMasterIPList()) == 1)` ,i think it is no need to judge it.   Code logic will protect the cluster have a master at least.

apply/processor/utils.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"github.com/sealerio/sealer/pkg/runtime"
+	"github.com/sealerio/sealer/pkg/runtime/k0s"
+	"github.com/sealerio/sealer/pkg/runtime/kubernetes"
+	"github.com/sealerio/sealer/pkg/runtime/kubernetes/kubeadm"
+	v2 "github.com/sealerio/sealer/types/api/v2"
+)
+
+func RuntimeChoose(rootfs string, cluster *v2.Cluster, config *kubeadm.KubeadmConfig) (runtime.Interface, error) {
```
allencloud, line 25: @starComingup Here I have a point to discuss with you. 

If runtime is a field of details processor instance, or belongs to a processor's function?

If runtime is a field of processor, then RuntimeChoose should be used in a processor's initialization, rather than set runtime in processor's function.

For example, I see 
```
type ScaleProcessor struct {
    runtime runtime.Interface
    xxx
}
```

And I see 
```
func (s *ScaleProcessor) PreProcess(cluster *v2.Cluster) error {
	runTime, err := RuntimeChoose(platform.DefaultMountClusterImageDir(cluster.Name), cluster, s.ClusterFile.GetKubeadmConfig())
	if err != nil {
		return fmt.Errorf("failed to init default runtime: %v", err)
	}
	s.Runtime = runTime
```
I feel quite strange. Why do you `RuntimeChoose` in ScaleProcessor and set ScaleProcessor's runtime filed, instead of set runtime type in `NewScaleProcessor`? Initialization in NewScaleProcessor could make code much more clear and more object-oriented. 

WDYT?

allencloud: In addition, please change the function name into `ChooseRuntime` or `GetRuntime` rather than `RuntimeChoose`.

apply/processor/utils.go
```diff
@@ -0,0 +1,39 @@
...
+		return nil, err
+	}
+	switch metadata.ClusterRuntime {
+	case runtime.K8s:
+		return kubernetes.NewDefaultRuntime(cluster, config)
+	case runtime.K0s:
+		return k0s.NewK0sRuntime(cluster)
+	// Todo case runtime.K3s:
+	default:
+		return kubernetes.NewDefaultRuntime(cluster, config)
```
allencloud, line 37: I wish to see more thinking details about the `default` case. From my point of view, if there is no metadata.ClusterRuntime in image rootfs details, we take it as k8s by default. Because in the original implementation of ClusterImage, there is no metadata.ClusterRuntime, and we have to make the new version of sealer compatible with the legacy ClusterImage format. For what is worth, I suggest to add a log to tell end-user that he is using a legacy version, and using new version of sealer and ClusterImage is encouraged.

Otherwise, I think you should add a line of explicit log at Line31 to show the ClusterRuntime info.

pkg/runtime/k0s/reset.go
```diff
@@ -0,0 +1,81 @@
...
+			if err := k.resetNode(node); err != nil {
+				return fmt.Errorf("failed to reset node %s: %v", node, err)
+			}
+			return nil
+		})
+	}
+	return eg.Wait()
+}
+
+func (k *Runtime) resetMasters(nodes []net.IP) error {
```
allencloud, line 50: Do we need to use a `eg, _ := errgroup.WithContext(context.Background())` in `resetMasters` as well, just like what you did in `resetNodes`? @starComingup

pkg/runtime/k0s/runtime.go
```diff
@@ -38,48 +41,60 @@ type Runtime struct {
 	RegConfig *registry.Config
 }
 
+var ForceDelete bool
```
allencloud, line 44: I have not see any assignment to this variable. Please double check it.

pkg/runtime/k0s/upgrade.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (k *Runtime) upgradeNodes(IPs []net.IP, binPath string) error {
+	var nodeCmds = []string{
+		fmt.Sprintf("chmod +x %s/*", binPath),
+		fmt.Sprintf("mv %s/* /usr/bin", binPath),
+		"k0s stop",
+		"k0s start",
+	}
+	var err error
+	for _, ip := range IPs {
```
allencloud, line 81: Do we need to use a eg, _ := errgroup.WithContext(context.Background()) here?

pkg/runtime/k0s/upgrade.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+	err = k.upgradeNodes(k.cluster.GetNodeIPList(), binPath)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (k *Runtime) upgradeMasters(IPs []net.IP, binPath string) error {
```
allencloud, line 50: Do we need to use a `eg, _ := errgroup.WithContext(context.Background())` here?

allencloud:
Left some comment, and wait for @starComingup 's feedback and update before merging.

apply/processor/utils.go
starComingup: I agree with add a cluster runtime info and hint end-user using new version ClusterImage.

apply/processor/utils.go
starComingup: > Here I have a point to discuss with you.
> 
> If runtime is a field of details processor instance, or belongs to a processor's function?
> 
> If runtime is a field of processor, then RuntimeChoose should be used in a processor's initialization, rather than set runtime in processor's function.
> 
> For example, I see

The old code set kubernetes.runtime here, so I just make this. But when I coded this, I also felt confused why origin logic set runtime in sub function. I will try to modify it!

pkg/runtime/k0s/reset.go
starComingup: I think we can do what you advice, then I will verify it.

pkg/runtime/k0s/runtime.go
starComingup: Using this in Line45

pkg/runtime/k0s/upgrade.go
starComingup: I don't think it can speed up too much, so I'll keep this logic for now.

pkg/runtime/k0s/upgrade.go
starComingup: keep this logic for now, If necessary, I will modify it.

starComingup:
All comment and code updated! Waiting for test.

allencloud (approved):

allencloud:
Thanks greatly for your patience during review. @starComingup I think your work could improve sealer quite a lot. Thanks again.

PR #1690, opened 2022-09-07
FIX: skip render the subcharts for sealer build.

### LINKED ISSUE #1689: sealer build by RenderHelmChart render subcharts
### What happen?

When create a kubefile like this:
```dockerfile
FROM kubernetes:v1.19.8
COPY apps-all-in-one charts
```
And the chart `apps-all-in-one` has dependencies(subcharts: app1,app2,etc.)
sealer will render every subcharts in the chart, this may cause the unwanted images or global tpl error:
image

### Relevant log output?

_No response_

The RenderHelmChart might skip render the subcharts.

_No response_

_No response_

{"gitVersion":"v0.8.6","gitCommit":"884513e","buildDate":"2022-07-12 02:58:54","goVersion":"go1.16.15","compiler":"gc","platform":"linux/amd64"}

CentOS 7

3.10.0-862.el7.x86_64

- Install tools:
- Others:

tgfree7:

Skip render the subcharts.

Try to fix #1689

### DIFF
--- a/build/buildimage/differ.go
+++ b/build/buildimage/differ.go
@@ -112,6 +112,7 @@ func parseChartImages(srcPath string) ([]string, error) {
 				return err
 			}
 			images = append(images, ima...)
+			return filepath.SkipDir
 		}
 		return nil
 	})


justadogistaken:
@tgfree7 @kakaZhou719 The render for subcharts should be skipped? I think I need more information.

tgfree7:
> @tgfree7 @kakaZhou719 The render for subcharts should be skipped? I think I need more information.

Render subcharts may list unwanted images, like this:
values.yaml for chart `apps-all-in-one`

```yaml
app1:
    Registry: release.com
    image: app1:v1_2022
```
and values.yaml for subcharts `app1` in the `apps-all-in-one`

```yaml
app1:
    Registry: test.com
    image: app1:v1_2022
```

In the case, render the subchart `app1` will list the unwanted images. `test.com:app1:v1_2022`

**And render subcharts may cause the tpl error, if `app1` call the tpl of _helper.tpl in the `apps-all-in-one`**

tgfree7:
> @tgfree7 If i'm not mistaken, subcharts has their own values? And of course the parent values could be passed to the subcharts. But what will happen if the user need the "test/com/app1:v1_2022"

Yes, subcharts has their own values.yaml. 
But when we release the chart `apps-all-in-one` and set the subchart `app1`(s Registry, images,etc.) by the parent chart values.yaml. this means we just need the image `release.com/app1:v1_2022` in this helm chart.
For the chart `apps-all-in-one`, we release the apps(app1,app2,etc) in one chart.

justadogistaken:
Okay, I understand what you mean. But in this case, we need to know that the (registry. images) for `app1` has been specified in parent values.yaml, Right? What if user assign the `app1` (registry. images) in subcharts only? I think it will igonre the image.

tgfree7:
> Okay, I understand what you mean. But in this case, we need to know that the (registry. images) for `app1` has been specified in parent values.yaml, Right? What if user assign the `app1` (registry. images) in subcharts only? I think it will igonre the image.

If user assign the `app1` (registry. images) in subcharts only, helm render will use the subcharts' setting.
The [subcharts_and_globals](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/) shows the details.

justadogistaken:
> > Okay, I understand what you mean. But in this case, we need to know that the (registry. images) for `app1` has been specified in parent values.yaml, Right? What if user assign the `app1` (registry. images) in subcharts only? I think it will igonre the image.
> 
> If user assign the `app1` (registry. images) in subcharts only, helm render will use the subcharts' setting. The [subcharts_and_globals](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/) shows the details.

Yes, helm will render, but the render logic in sealer will do the same thing? Could you help confirm this? Write a test for this part will be perfect.

tgfree7:
> > > Okay, I understand what you mean. But in this case, we need to know that the (registry. images) for `app1` has been specified in parent values.yaml, Right? What if user assign the `app1` (registry. images) in subcharts only? I think it will igonre the image.
> > 
> > 
> > If user assign the `app1` (registry. images) in subcharts only, helm render will use the subcharts' setting. The [subcharts_and_globals](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/) shows the details.
> 
> Yes, helm will render, but the render logic in sealer will do the same thing? Could you help confirm this? Write a test for this part will be perfect.

Yes, i confirm that RenderHelmChart will render in the same way,and i will write a test for the RenderHelmChart.

justadogistaken:
@tgfree7 Thanks for your work. I think you can create test samples(subchart included and excluded). And call the function of sealer, check the expected content.

tgfree7:
> @tgfree7 Thanks for your work. I think you can create test samples(subchart included and excluded). And call the function of sealer, check the expected content.

In the test case, where the chart should i store? in the sealer's path `test` ? I reivew the old test case, the chart `./test/alpine` are not found.

justadogistaken:
@tgfree7 You can create a `testfile` dir under the build or buildimage.

tgfree7:
> @tgfree7 You can create a `testfile` dir under the build or buildimage.

test case added.

build/layerutils/charts/helm_test.go
```diff
@@ -14,6 +14,12 @@
 
 package charts
 
+import (
+	"reflect"
+	"sort"
+	"testing"
+)
+
 /* func TestPackageHelmChart(t *testing.T) {
```
justadogistaken, line 23: Could you help remove the unused function？

tgfree7: this function called in the case TestGetImageList, did  you mean TestRenderHelmChart and  TestPackageHelmChart?

justadogistaken: Yes, I think no one use these functions to do test. They are archived.

tgfree7: ok.

justadogistaken:
There is license problem.
`make install-addlicense && make filelicense`
Could you plz try this command?
And Probably the `install-addlicense` will fail, in that case, you can try install it yourself, and do make filelicense.

tgfree7:
> make install-addlicense



> There is license problem. `make install-addlicense && make filelicense` Could you plz try this command? And Probably the `install-addlicense` will fail, in that case, you can try install it yourself, and do make filelicense.

ok, license added.

tgfree7:
> There is license problem. `make install-addlicense && make filelicense` Could you plz try this command? And Probably the `install-addlicense` will fail, in that case, you can try install it yourself, and do make filelicense.

ok, license added.

justadogistaken:
LGTM
@kakaZhou719 Help check this work.

justadogistaken (approved):

PR #1698, opened 2022-09-09
Fix: overlay umount error by mountpoint

### LINKED ISSUE #1697: sealer prune failed when prune build tmp
### What happen?

When sealer image failed, try to prune the data failed by command `sealer prune`.

### Relevant log output?

```Shell
# sealer prune
start to prune image db ... 
start to prune layer ... 
3ca5ae9d0b24deabce12213b927664dc692f389dc2e826a1f301df4a4ba6a873 deleted
3ca5ae9d0b24deabce12213b927664dc692f389dc2e826a1f301df4a4ba6a873 deleted
Usage:
  sealer prune [flags]

Examples:
$ sealer prune
start to prune image db ...
start to prune layer ...
start to prune build tmp ...


Flags:
  -h, --help   help for prune

Global Flags:
      --color string               set the log color mode, the possible values can be [never always] (default "always")
      --config string              config file of sealer tool (default is $HOME/.sealer.json)
  -d, --debug                      turn on debug mode
      --hide-path                  hide the log path
      --hide-time                  hide the log time
      --log-to-file                write log message to disk
  -q, --quiet                      silence the usage when fail
      --remote-logger-url string   remote logger url, if not empty, will send log to this url
      --task-name string           task name which will embedded in the remote logger header, only valid when --remote-logger-url is set

2022-09-09 03:03:52 [ERROR] [root.go:70] sealer-v0.8.6: failed to umount target /var/lib/sealer/tmp/.DTmp-1667720254: invalid argument
```

pruned succeeds.

```YAML
create a sealer-build failed. and make sure there is a overlay mount created by sealer. like this:

]# df -h
Filesystem               Size  Used Avail Use% Mounted on
/dev/mapper/centos-root   40G  9.6G   31G  24% /
devtmpfs                 1.9G     0  1.9G   0% /dev
tmpfs                    1.9G     0  1.9G   0% /dev/shm
tmpfs                    1.9G  8.6M  1.9G   1% /run
tmpfs                    1.9G     0  1.9G   0% /sys/fs/cgroup
/dev/mapper/centos-home   20G  259M   20G   2% /home
/dev/sda1               1014M  142M  873M  14% /boot
tmpfs                    379M     0  379M   0% /run/user/0
overlay                   40G  9.6G   31G  24% /var/lib/sealer/tmp/.DTmp-4069575571
```

_No response_

{"gitVersion":"v0.8.6","gitCommit":"884513e","buildDate":"2022-07-12 02:58:54","goVersion":"go1.16.15","compiler":"gc","platform":"linux/amd64"}

CentOS 7

3.10.0-862.el7.x86_64

- Install tools:
- Others:

tgfree7:

Umount the tmp overlay device by device path not mountpoint.

In the sealer (<=v0.8.6), sealer prune will prune build-tmp-data by umount and remove, but umount overlay type monuntpoint will failed. maybe we should just umount the device name not mountpoint.

I find out, in the master branch, the command `prune` has dropped. should i fix in the 0.8.6?

try to fix #1697

### DIFF
--- a/utils/mount/mount_service.go
+++ b/utils/mount/mount_service.go
@@ -19,12 +19,12 @@ import (
 	"net"
 	"strings"
 
+	"github.com/moby/sys/mountinfo"
+	"github.com/sealerio/sealer/utils/ssh"
 	"github.com/shirou/gopsutil/disk"
 	"github.com/sirupsen/logrus"
 
-	"github.com/sealerio/sealer/utils/exec"
 	"github.com/sealerio/sealer/utils/os/fs"
-	"github.com/sealerio/sealer/utils/ssh"
 	strUtils "github.com/sealerio/sealer/utils/strings"
 )
 
@@ -120,6 +120,18 @@ func NewMountService(target, upper string, lowLayers []string) (Service, error) 
 	}, nil
 }
 
+func GetDirNameListInDir(dir string) ([]string, error) {
+	var dirs []string
+	infos, err := mountinfo.GetMounts(mountinfo.PrefixFilter(dir))
+	if err != nil {
+		return dirs, err
+	}
+	for _, info := range infos {
+		dirs = append(dirs, info.Mountpoint)
+	}
+	return dirs, nil
+}
+
 //NewMountServiceByTarget will filter file system by target,if not existed,return false.
 func NewMountServiceByTarget(target string) Service {
 	mounted, info := GetMountDetails(target)
@@ -141,12 +153,11 @@ type Info struct {
 }
 
 func GetMountDetails(target string) (bool, *Info) {
-	cmd := fmt.Sprintf("mount | grep %s", target)
-	result, err := exec.RunSimpleCmd(cmd)
+	infos, err := mountinfo.GetMounts(mountinfo.SingleEntryFilter(target))
 	if err != nil {
 		return false, nil
 	}
-	return mountCmdResultSplit(result, target)
+	return mountCmdResultSplit(infos[0].VFSOptions, target)
 }
 
 func GetRemoteMountDetails(s ssh.Interface, ip net.IP, target string) (bool, *Info) {
--- a/pkg/prune/build_pruner.go
+++ b/pkg/prune/build_pruner.go
@@ -17,7 +17,6 @@ package prune
 import (
 	"github.com/sealerio/sealer/common"
 	"github.com/sealerio/sealer/utils/mount"
-	osi "github.com/sealerio/sealer/utils/os"
 )
 
 type buildPrune struct {
@@ -33,10 +32,7 @@ func NewBuildPrune() Pruner {
 func (b buildPrune) Select() ([]string, error) {
 	var pruneList []string
 	// umount all tmp dir, and delete it
-	pruneUnits, err := osi.GetDirNameListInDir(b.pruneRootDir, osi.FilterOptions{
-		All:          true,
-		WithFullPath: true,
-	})
+	pruneUnits, err := mount.GetDirNameListInDir(b.pruneRootDir)
 	if err != nil {
 		return pruneList, err
 	}


justadogistaken:
@tgfree7 For the version newer than 0.8.6. We discard the original db for layer and image.
We implement transactions about image based on `buildah`.
You can check imageengine out.
And could you please help us improve the prune(by invoking the function of buildah)?
The test image is `registry.cn-qingdao.aliyuncs.com/sealer-io/kubernetes:v1.19.8.test`
And you can also build some local image on your own to test prune.

tgfree7:
> @tgfree7 For the version newer than 0.8.6. We discard the original db for layer and image. We implement transactions about image based on `buildah`. You can check imageengine out. And could you please help us improve the prune(by invoking the function of buildah)? The test image is `registry.cn-qingdao.aliyuncs.com/sealer-io/kubernetes:v1.19.8.test` And you can also build some local image on your own to test prune.

ok,  i will check the imageengine out, and try to improve the prune.

justadogistaken:
@tgfree7 
I checked the caller. And I found we retrive the mount info by "mount".
And I think there are some libraries supporting that ability.
I found this lib [mountinfo](https://pkg.go.dev/github.com/moby/sys/mountinfo), But I'm not so sure.
Could you help make it better? I really appreciate it.

justadogistaken:
@tgfree7 prune can be completed in another pr.

tgfree7:
> @tgfree7 I checked the caller. And I found we retrive the mount info by "mount". And I think there are some libraries supporting that ability. I found this lib [mountinfo](https://pkg.go.dev/github.com/moby/sys/mountinfo), But I'm not so sure. Could you help make it better? I really appreciate it.

i will the lib act as expected.

tgfree7:
> @tgfree7 I checked the caller. And I found we retrive the mount info by "mount". And I think there are some libraries supporting that ability. I found this lib [mountinfo](https://pkg.go.dev/github.com/moby/sys/mountinfo), But I'm not so sure. Could you help make it better? I really appreciate it.

the lib [mountinfo](https://pkg.go.dev/github.com/moby/sys/mountinfo) can do :
* check a specified path is mountpoint or not;
* list all mounts for a base path(like `/var/lib/sealer/tmp/`);

But can not return the mountinfo in sealer-style(show the lower,upper dir), still need a func like `mountCmdResultSplit` to get overlay info

justadogistaken:
@tgfree7 
Ok, let's keep the current way processing it.
I'm prefer to use regular expression to handle the string.
For example:
```
overlay on /home/ecs-assist-user/test/merged type overlay (rw,relatime,lowerdir=lower:lower2/,upperdir=upper,workdir=work)
```
* workdir=[^,)]+
* lowerdir=[^,)]+
* upperdir=[^,)]+

utils/mount/mount_service.go
```diff
@@ -123,7 +123,7 @@ func NewMountService(target, upper string, lowLayers []string) (Service, error)
 //NewMountServiceByTarget will filter file system by target,if not existed,return false.
 func NewMountServiceByTarget(target string) Service {
 	mounted, info := GetMountDetails(target)
-	if !mounted {
+	if !mounted || !strings.Contains(info.Work, target) {
```
justadogistaken, line 126: I'm confused that why workdir will contains the target? I guess they are not relevant?

tgfree7:
> 

maybe we can try to use `mountinfo.GetMounts(mountinfo.PrefixFilter(dir))` to list correct mountpoints 
and get info (need regular expression to handle the string) by `mountinfo.GetMounts(mountinfo.SingleEntryFilter("$base_dir"))`?
in this way, func GetMounts can filter the mountpoints.
Is it worth to import a package?

utils/mount/mount_service.go
tgfree7: Now sealer will list all dirs in the /var/lib/sealer/tmp like this
image

And sealer mount the path like this:
```golang
func (o *Overlay2) Mount(target string, upperLayer string, layers ...string) error {
	if target == "" {
		return fmt.Errorf("target cannot be empty")
	}
	if len(layers) == 0 {
		return fmt.Errorf("layers cannot be empty")
	}
	workdir := path.Join(target, "work")
	if err := os.MkdirAll(workdir, os.ModePerm); err != nil {
		return fmt.Errorf("failed to create workdir: %v", err)
	}
	var err error
	defer func() {
		if err != nil {
			_ = os.RemoveAll(workdir)
		}
	}()

	var indexOff string
	// figure out whether "index=off" option is recognized by the kernel
	_, err = os.Stat("/sys/module/overlay/parameters/index")
	switch {
	case err == nil:
		indexOff = "index=off,"
	case os.IsNotExist(err):
		// old kernel, no index -- do nothing
	default:
		logrus.Warnf("Unable to detect whether overlay kernel module supports index parameter: %s", err)
	}

	mountData := fmt.Sprintf("%slowerdir=%s,upperdir=%s,workdir=%s", indexOff, strings.Join(strUtils.Reverse(layers), ":"), upperLayer, workdir)
	logrus.Debugf("mount data : %s", mountData)
	if err = mount("overlay", target, "overlay", 0, mountData); err != nil {
		return fmt.Errorf("failed to create overlay mount to %s: %v", target, err)
	}
	return nil
}
```

justadogistaken:
> > 
> 
> maybe we can try to use `mountinfo.GetMounts(mountinfo.PrefixFilter(dir))` to list correct mountpoints and get info (need regular expression to handle the string) by `mountinfo.GetMounts(mountinfo.SingleEntryFilter("$base_dir"))`? in this way, func GetMounts can filter the mountpoints. Is it worth to import a package?

IMO, It's worthy.

tgfree7:
> > > 
> > 
> > 
> > maybe we can try to use `mountinfo.GetMounts(mountinfo.PrefixFilter(dir))` to list correct mountpoints and get info (need regular expression to handle the string) by `mountinfo.GetMounts(mountinfo.SingleEntryFilter("$base_dir"))`? in this way, func GetMounts can filter the mountpoints. Is it worth to import a package?
> 
> IMO, It's worthy.

ok

kakaZhou719 (approved):
LGTM

justadogistaken:
@tgfree7 Do you test this?

tgfree7:
> @tgfree7 Do you test this?

i test this on version 0.8.6

justadogistaken:
@tgfree7 
Could you help check the main branch.
We currently will use buildah to build image.
In this way, it will generate lots of intermediate image, and buildah containers.
I think we could implement prune to remove the intermediate images and buildah containers.
You can install buildah to list all images to find the intermediate images and buildah containers.

tgfree7:
> @tgfree7 Could you help check the main branch. We currently will use buildah to build image. In this way, it will generate lots of intermediate image, and buildah containers. I think we could implement prune to remove the intermediate images and buildah containers. You can install buildah to list all images to find the intermediate images and buildah containers.

OK this will be in the next pr.

PR #1700, opened 2022-09-11
feat: relaxing overlay mount restrictions

### LINKED ISSUE #1547: 放宽overlay挂载限制
问题描述
目前sealer build image时会检测是否支持overlay文件系统。当前默认通过modprobe overlay去检测，我发现在容器环境和wsl2都会导致最终保存失败(能够保存，但是拉取的镜像丢失，算作保存失败)。

解决方案：
去除overlay2.go中modprobe检测，仅使用/proc/filesystems去检测。并希望能够在不支持overlay的情况下，通过日志在控制台上显示。
https://github.com/sealerio/sealer/blob/cf9bdb1667d6d519aa0af63037999d73646dcbb4/utils/mount/overlay2.go#L49

YTGhost:
relaxing overlay mount restrictions

Fixes #1547

### DIFF
--- a/utils/mount/overlay2.go
+++ b/utils/mount/overlay2.go
@@ -18,10 +18,8 @@
 package mount
 
 import (
-	"bufio"
 	"fmt"
 	"os"
-	"os/exec"
 	"path"
 	"strings"
 	"syscall"
@@ -46,26 +44,17 @@ func NewMountDriver() Interface {
 	return &Default{}
 }
 
-func supportsOverlay() bool {
-	if err := exec.Command("modprobe", "overlay").Run(); err != nil {
-		return false
-	}
-	f, err := os.Open("/proc/filesystems")
+func supportsOverlay() (flag bool) {
+	content, err := os.ReadFile("/proc/filesystems")
 	if err != nil {
 		return false
 	}
 	defer func() {
-		if err := f.Close(); err != nil {
-			logrus.Errorf("failed to close file: %v", err)
+		if !flag {
+			logrus.Warn("The current environment does not support overlay")
 		}
 	}()
-	s := bufio.NewScanner(f)
-	for s.Scan() {
-		if s.Text() == "nodev\toverlay" {
-			return true
-		}
-	}
-	return false
+	return strings.Contains(string(content), "overlay")
 }
 
 // Mount using overlay2 to merged layer files


justadogistaken:
I'm considering if `modprobe overlay` failed, why do we allow the process to keep on. Why no just throw the error? I'm afraid that relaxing the restriction will incur more problems. And I think `modprobe overlay` is enough for now. If there are users need to relax the restriction. We can provide a env or option for user.
@YTGhost @ningpengtao-coder

utils/mount/overlay2.go
```diff
@@ -46,15 +45,15 @@ func NewMountDriver() Interface {
...
-		return false
-	}
+func supportsOverlay() (flag bool) {
 	f, err := os.Open("/proc/filesystems")
 	if err != nil {
 		return false
 	}
 	defer func() {
+		if !flag {
+			logrus.Infof("The current environment does not support overlay")
```
justadogistaken, line 55: Maybe warn would be better

YTGhost: Yep, Thx!

YTGhost: > I'm considering if `modprobe overlay` failed, why do we allow the process to keep on. Why no just throw the error? I'm afraid that relaxing the restriction will incur more problems. And I think `modprobe overlay` is enough for now. If there are users need to relax the restriction. We can provide a env or option for user. @YTGhost @ningpengtao-coder

I think using `/proc/filesystems` to detect supported filesystems seems to be sufficient in a linux environment? In wsl and container environments, `modprobe overlay` may not be supported for use.

YTGhost:
> I'm considering if `modprobe overlay` failed, why do we allow the process to keep on. Why no just throw the error? I'm afraid that relaxing the restriction will incur more problems. And I think `modprobe overlay` is enough for now. If there are users need to relax the restriction. We can provide a env or option for user. @YTGhost @ningpengtao-coder

I think using `/proc/filesystems` to detect supported filesystems seems to be sufficient in a linux environment? In wsl and container environments, `modprobe overlay` may not be supported for use.

justadogistaken:
@YTGhost Ok, I understand that. Thanks for this work.

utils/mount/overlay2.go
```diff
@@ -46,15 +45,15 @@ func NewMountDriver() Interface {
...
-		return false
-	}
+func supportsOverlay() (flag bool) {
 	f, err := os.Open("/proc/filesystems")
 	if err != nil {
 		return false
 	}
 	defer func() {
+		if !flag {
+			logrus.Warnf("The current environment does not support overlay")
```
justadogistaken, line 55: Change it to Warn insteand of Warnf.

YTGhost: Done

utils/mount/overlay2.go
```diff
@@ -46,15 +45,15 @@ func NewMountDriver() Interface {
 	return &Default{}
 }
 
-func supportsOverlay() bool {
```
justadogistaken, line 49: Hang on a second.
I think there are two condition to detect the overlay support.
"modprobe overlay" || "/proc/filesystems".
I'm just afraid that some of cases that there will be false in "/proc/filesystems" case.

justadogistaken:
Sorry, I misunderstand this issue.
So it could happen that "mobprobe overlay" returns true, but nothing in the "/proc/filesystems", am I right?

justadogistaken:
There is a function "checkOverlaySupported" in sealer:main.
You can check this out. Maybe we can merge the current function and the "checkOverlaySupported". After this work committed.

justadogistaken:
@YTGhost 
```
	s := bufio.NewScanner(f)
	for s.Scan() {
		if s.Text() == "nodev\toverlay" {
			return true
		}
	}
	return false
```
For this part. I guess we could read them all in.
And check if "overlay" exist?
Should "nodev\t" be included?

justadogistaken (approved):

justadogistaken:
Please change `if s.Text() == "nodev\toverlay"` to strings.Containes, then I will merge the pr.
And we don't have to scan(). Just read the file content in.

YTGhost:
> Sorry, I misunderstand this issue. So it could happen that "mobprobe overlay" returns true, but nothing in the "/proc/filesystems", am I right?

It could happen that "mobprobe overlay" returns false, but exist in the "/proc/filesystems"

YTGhost:
> There is a function "checkOverlaySupported" in sealer:main. You can check this out. Maybe we can merge the current function and the "checkOverlaySupported". After this work committed.

Yep! we can try to merge them

justadogistaken:
@YTGhost Okay. We can complete current work first. And file a another pr for merging two functions.

utils/mount/overlay2.go
```diff
@@ -46,24 +44,18 @@ func NewMountDriver() Interface {
...
+		if !flag {
+			logrus.Warn("The current environment does not support overlay")
 		}
 	}()
-	s := bufio.NewScanner(f)
-	for s.Scan() {
-		if s.Text() == "nodev\toverlay" {
-			return true
-		}
+	if strings.Contains(string(content), "nodev\toverlay") {
```
justadogistaken, line 57: “nodev\toverlay” => "overlay"

YTGhost: Done

utils/mount/overlay2.go
```diff
@@ -46,24 +44,18 @@ func NewMountDriver() Interface {
...
+		if !flag {
+			logrus.Warn("The current environment does not support overlay")
 		}
 	}()
-	s := bufio.NewScanner(f)
-	for s.Scan() {
-		if s.Text() == "nodev\toverlay" {
-			return true
-		}
+	if strings.Contains(string(content), "overlay") {
```
justadogistaken, line 57: return strings.Contains(string(content), "overlay")

YTGhost: Done

justadogistaken:
LGTM! @YTGhost Thanks.

justadogistaken:
I just created a ecs.
I do as follows:
1. cat /proc/filesystems, no overlay.
2. modprobe overlay.
3. cat /proc/filesystems, overlay appears.

So I recommend that add a line to the first line of this function.
But we don't have to catch the return.
```
_ = exec.Command("modprobe", "overlay").Run()
```

YTGhost:
> I just created a ecs. I do as follows:
> 
> 1. cat /proc/filesystems, no overlay.
> 2. modprobe overlay.
> 3. cat /proc/filesystems, overlay appears.
> 
> So I recommend that add a line to the first line of this function. But we don't have to catch the return.
> 
> ```
> _ = exec.Command("modprobe", "overlay").Run()
> ```

Yep, I think we should add it to make sure the module is loaded？
Should I create a new branch to modify it?

justadogistaken:
@YTGhost 
Yes.
File a new pr to fix this, plz.

PR #1716, opened 2022-09-19
optimize the buildah options

justadogistaken:
There are general options for cmds could be put into the engine. Instead of initialize one for every command.
And this could help to fix redefine problem.
@VinceCui @kakaZhou719 FYI.


VinceCui (approved):
LGTM

PR #1722, opened 2022-09-20
fix:SaveAll funcation

Stevent-fei:
1. Fix saveAll function
2. Add unit tests


VinceCui (approved):
LGTM

PR #1727, opened 2022-09-22
code optimization for sealer CLI

kakaZhou719:
1. polish CLI: delete unused flag, add example for join, delete ,run;
2. let all master could access api server.
3. split clean logic to kube reset and container runtime uninstall.
4. return error fast if user not set required flag Or args.


VinceCui:
Very important PR, thx.

PR #1728, opened 2022-09-22
move BuildRootfs content to CreateWorkingContainer

jsparter:
Signed-off-by: jsparter <1084237763@qq.com>
The actual action of the current `BuildRootfs` is simply to create a Buildah's working-container and then mount it to a directory. By modifying the contents of the working-container, Buildah finally modify the image. This is a step of Buildah's workflow. Therefore, `CreateWorkingContainer` is more closely matches the description of this action.
Maybe `BuildRootfs` needs to clarify its responsibility


pkg/imageengine/buildah/build_rootfs.go
```diff
@@ -15,36 +15,11 @@
...
 	"github.com/sealerio/sealer/pkg/define/options"
-
-	"os"
 )
 
-// BuildRootfs will make an image rootfs under /var/lib/containers/storage
-// And then link it to the DestDir
-// And remember to call RemoveContainer to remove the link and remove the container(umount rootfs) manually.
+// BuildRootfs need to clarify responsibility
 func (engine *Engine) BuildRootfs(opts *options.BuildRootfsOptions) (containerID string, err error) {
```
justadogistaken, line 22: Why not delete "BuildRootfs"

jsparter: IMO, `BuildRootfs` sounds like it is building rootfs, which is what end users might use when building their own rootfs. I'm not sure whether we need the func in the future. Or, as you annotated before, `BuildRootfs` should be upper to image engine

justadogistaken: Let's remove it first. We will add it when we need it.

jsparter: removed

justadogistaken (approved):
LGTM

PR #1729, opened 2022-09-22
feat: set default image pull policy as always

### LINKED ISSUE #1453: Support set image pull policy in sealer pull, now maybe IfNotPresent
## Issue Description

Type: *feature request*

Support set image pull policy in sealer pull, now maybe IfNotPresent #1453

I think maybe should support Always, and set it as default.

YTGhost:
Signed-off-by: HIHIA <283304489@qq.com>

set default image pull policy as always, and add ifnewer in usage

Fixes #1453

### DIFF
--- a/cmd/sealer/cmd/image/pull.go
+++ b/cmd/sealer/cmd/image/pull.go
@@ -46,7 +46,7 @@ func NewPullCmd() *cobra.Command {
 	pullOpts = &options.PullOptions{}
 	pullCmd.Flags().StringVar(&pullOpts.Platform, "platform", parse.DefaultPlatform(), "prefer OS/ARCH instead of the current operating system and architecture for choosing images")
 	pullCmd.Flags().BoolVar(&pullOpts.TLSVerify, "tls-verify", true, "require HTTPS and verify certificates when accessing the registry. TLS verification cannot be used when talking to an insecure registry.")
-	pullCmd.Flags().StringVar(&pullOpts.PullPolicy, "policy", "missing", "missing, always, or never.")
+	pullCmd.Flags().StringVar(&pullOpts.PullPolicy, "policy", "always", "missing, always, ifnewer or never.")
 	pullCmd.Flags().BoolVarP(&pullOpts.Quiet, "quiet", "q", false, "don't output progress information when pulling images")
 	return pullCmd
 }


cmd/sealer/cmd/image/pull.go
```diff
@@ -46,7 +46,7 @@ func NewPullCmd() *cobra.Command {
 	pullOpts = &options.PullOptions{}
 	pullCmd.Flags().StringVar(&pullOpts.Platform, "platform", parse.DefaultPlatform(), "prefer OS/ARCH instead of the current operating system and architecture for choosing images")
 	pullCmd.Flags().BoolVar(&pullOpts.TLSVerify, "tls-verify", true, "require HTTPS and verify certificates when accessing the registry. TLS verification cannot be used when talking to an insecure registry.")
-	pullCmd.Flags().StringVar(&pullOpts.PullPolicy, "policy", "missing", "missing, always, or never.")
+	pullCmd.Flags().StringVar(&pullOpts.PullPolicy, "policy", "always", "missing, always, newer or never.")
```
justadogistaken, line 49: There is no newer syntax currently.
It should be "ifnewer".
I think "ifnewer" should be the default option.
@VinceCui

YTGhost: > There is no newer syntax currently. It should be "ifnewer". I think "ifnewer" should be the default option. @VinceCui

Yes I noticed this at first too, it does use "ifnewer" in `define/pull.go`. But I looked at buildah's source code about pull, it will execute the following ParsePullPolicy function:
```go
// ParsePullPolicy parses the string into a pull policy.
//
// Supported string values are:
// * "always"  <-> PullPolicyAlways
// * "missing" <-> PullPolicyMissing (also "ifnotpresent" and "")
// * "newer"   <-> PullPolicyNewer (also "ifnewer")
// * "never"   <-> PullPolicyNever
func ParsePullPolicy(s string) (PullPolicy, error) {
	switch s {
	case "always", "Always":
		return PullPolicyAlways, nil
	case "missing", "Missing", "ifnotpresent", "IfNotPresent", "":
		return PullPolicyMissing, nil
	case "newer", "Newer", "ifnewer", "IfNewer":
		return PullPolicyNewer, nil
	case "never", "Never":
		return PullPolicyNever, nil
	default:
		return PullPolicyUnsupported, errors.Errorf("unsupported pull policy %q", s)
	}
}
```
And I checked buildah's documentation on pull https://github.com/containers/buildah/blob/main/docs/buildah-pull.1.md
It use newer, so I think we should be consistent with buildah?
```
--policy=always|missing|never|newer

Pull image policy. The default is missing.

always: Always pull the image and throw an error if the pull fails.
missing: Pull the image only if it could not be found in the local containers storage. Throw an error if no image could be found and the pull fails.
never: Never pull the image but use the one from the local containers storage. Throw an error if no image could be found.
newer: Pull if the image on the registry is newer than the one in the local containers storage. An image is considered to be newer when the digests are different. Comparing the time stamps is prone to errors. Pull errors are suppressed if a local image was found.
```

justadogistaken: Great.
"newer" is much easier.
But I think "newer" should be the default option.

justadogistaken: @YTGhost 
You can check "PolicyMap" in the pull function.
It could not support newer(the low-level lib supports it.)
Maybe "ifnewer" is good enough.

YTGhost: > @YTGhost You can check "PolicyMap" in the pull function. It could not support newer(the low-level lib supports it.) Maybe "ifnewer" is good enough.

Yes, my mistake.

justadogistaken (approved):
LGTM

PR #1744, opened 2022-09-28
Fix： the format for saving the cluster IP address

Stevent-fei:
fix the format for saving the cluster IP address
eg:
image


VinceCui (commented):
Good code is clean.

PR #1756, opened 2022-09-30
vendor kubeadm/v1beta2 rather than directly import this golang file

zhy76:
vendor kubeadm/v1beta2 rather than directly import this golang file

Fixes #1754 
### Describe how you did it
I deleted the kubeadm/v1beta2 and vendor kubeadm/v1beta2 rather than directly import this golang file,then I modified the other affected files

### DIFF
--- a/pkg/runtime/k0s/delete_masters.go
+++ b/pkg/runtime/k0s/delete_masters.go
@@ -61,7 +61,7 @@ func (k *Runtime) deleteMaster(master net.IP) error {
 	STEP7: delete node though k8s client
 	*/
 	remoteCleanCmd := []string{"k0s stop",
-		fmt.Sprintf("k0s reset --config %s --cri-socket %s", DefaultK0sConfigPath, ExternalCRI),
+		fmt.Sprintf("k0s reset --cri-socket %s", ExternalCRI),
 		"rm -rf /etc/k0s/",
 		fmt.Sprintf("sed -i \"/%s/d\" /etc/hosts", SeaHub),
 		fmt.Sprintf("sed -i \"/%s/d\" /etc/hosts", k.RegConfig.Domain),
--- a/pkg/runtime/k0s/delete_nodes.go
+++ b/pkg/runtime/k0s/delete_nodes.go
@@ -60,7 +60,7 @@ func (k *Runtime) deleteNode(node net.IP) error {
 	STEP7: delete node though k8s client
 	*/
 	remoteCleanCmds := []string{"k0s stop",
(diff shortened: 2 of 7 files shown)


jsparter:
I noticed that #1743  just means vendor to import files in need. But your `go.mod` has changed a lot, and cause lots of vendor file changed. Over the last few PR changed `go.mod` too, I'm confused is this nessesary?

allencloud:
@zhy76 Thanks for your great work. I have some confusion that why you updated so many vendored packages. From my point of view, just vendor kubeadm/v1beta2, there will not be so much update for other vendored packages. Could you explain a little bit for my question?

zhy76:
Ok,I excute the "go mod vendor" command and it updated so many vendored packages.This is my first pr,I'am sorry for this problem and I will fix this and try again,thanks for your advice!

starComingup:
> @zhy76 Thanks for your great work. I have some confusion that why you updated so many vendored packages. From my point of view, just vendor kubeadm/v1beta2, there will not be so much update for other vendored packages. Could you explain a little bit for my question?

@allencloud I have experienced this problem, It seems using different go version, when use go mod, it will update some vendor file.
@zhy76 IMO, using kubeadm/v1beta2 will less dependency with k8s vendor. Because import vendor will may introduce more useless code, but I support the way with your opinion. Just try to focus on resolving `go.mod` file.

allencloud:
> Ok,I excute the "go mod vendor" command and it updated so many vendored packages.This is my first pr,I'am sorry for this problem and I will fix this and try again,thanks for your advice!

Oh, it is not your fault. You do not need to apologize. Everyone has his first time to contribute. Everyone is bettering from his first step. I failed to do vendoring kubeadm/v1beta2. Actually we admire your first step so much. No matter it works or not, we pay more attention on the sharing, questioning and feedback. @zhy76

go.mod
```diff
@@ -240,3 +248,57 @@ replace (
 	golang.org/x/net => golang.org/x/net v0.0.0-20210510120150-4163338589ed
 	golang.org/x/sys => golang.org/x/sys v0.0.0-20220114195835-da31bd327af9
 )
+
+replace k8s.io/api => k8s.io/api v0.25.2
```
justadogistaken, line 252: The version is same with what you import above.
I guess there is no need to do replace here.

jsparter:
> ### Does this pull request fix one issue?
> Fixes #1754
Plus, refer the issue directly is much better than refer a closed PR

go.mod
zhy76: If not replace,it will cause this error:
image
This is because in Kubernetes go.mod, you'll find the following code:
image
In the Kubernetes repository, the packages listed under the require commands are marked as v0.0.0

PR #1762, opened 2022-10-09
add app installer to sealer run

kakaZhou719:
1. add app installer to sealer run after runtime reconstructing
2. rename  package name from `guest` to `application `.


justadogistaken:
May I ask that does this work needs to be cooperated with others? 
If don't. we'd better not push branch into official sealer repository next time.

PR #1764, opened 2022-10-09
add app installer to sealer run

kakaZhou719:

1. add app installer to sealer run after runtime reconstructing
2. rename package name from `guest` to `application` .
3. bugfix for kubeadm config vendor files


cmd/sealer/cmd/cluster/join.go
```diff
@@ -19,6 +19,9 @@ import (
 	"io/ioutil"
 	"path/filepath"
 
+	imagecommon "github.com/sealerio/sealer/pkg/define/options"
```
Starnop, line 22: usually, there should be three groups separated by blank for go-imports
1. the golang official package
2. the package for the current project
3. the others

pkg/cluster-runtime/installer.go
```diff
@@ -118,71 +124,126 @@ func getWorkerIPList(infraDriver infradriver.InfraDriver) []net.IP {
...
 
-func (i *Installer) Install() (registry.Driver, runtime.Driver, error) {
-	master0 := i.infraDriver.GetHostIPListByRole(common.MASTER)[0]
-	masters := i.infraDriver.GetHostIPListByRole(common.MASTER)
-	workers := getWorkerIPList(i.infraDriver)
-	all := append(masters, workers...)
+func (i *Installer) Install() error {
+	var (
+		masters    = i.infraDriver.GetHostIPListByRole(common.MASTER)
+		master0    = masters[0]
```
Starnop, line 130: what if the length of masters is equal with zero?

pkg/cluster-runtime/installer.go
```diff
@@ -118,71 +124,126 @@ func getWorkerIPList(infraDriver infradriver.InfraDriver) []net.IP {
...
+		workers    = getWorkerIPList(i.infraDriver)
+		all        = append(masters, workers...)
+		image      = i.infraDriver.GetClusterImageName()
+		launchCmds = i.infraDriver.GetClusterLaunchCmds()
```
Starnop, line 134: if these codes not only for install cluster,  it's prefer to use `clusterLaunchCmds` instead of `launchCmds`

and so do with image

pkg/cluster-runtime/installer.go
```diff
@@ -118,71 +124,126 @@ func getWorkerIPList(infraDriver infradriver.InfraDriver) []net.IP {
...
-	if err != nil {
-		return nil, nil, err
-	}
+	return nil
+}
 
-	if err := i.runClusterHook(master0, PostInstallCluster); err != nil {
-		return nil, nil, err
-	}
+func (i *Installer) installApp(master0 net.IP, cmds []string) error {
```
Starnop, line 226: what's the reason for the master0 IP is needed here for install application

In fact, we should allow the application image to be deployed on either node

pkg/cluster-runtime/installer.go
```diff
@@ -118,71 +124,126 @@ func getWorkerIPList(infraDriver infradriver.InfraDriver) []net.IP {
...
+		err = i.installKubeCluster(all)
+		if err != nil {
+			return fmt.Errorf("failed to install cluster: %s", err)
+		}
+
+		err = i.installApp(master0, cmds)
+		if err != nil {
+			return fmt.Errorf("failed to install application: %s", err)
+		}
+	} else {
```
Starnop, line 172: Where will the `CMDS` defined in the old image be executed?

pkg/imagedistributor/mount.go
```diff
@@ -104,11 +104,7 @@ func (c ImagerMounter) Umount(imageMountInfo map[string]string) error {
 	return nil
 }
 
-func NewImageMounter(hostsPlatform map[v1.Platform][]net.IP) (*ImagerMounter, error) {
-	imageEngine, err := imageengine.NewImageEngine(imagecommon.EngineGlobalConfigurations{})
-	if err != nil {
-		return nil, err
-	}
+func NewImageMounter(imageEngine imageengine.Interface, hostsPlatform map[v1.Platform][]net.IP) (*ImagerMounter, error) {
```
Starnop, line 107: nice job 👍 

Make each method as atomized as possible, and we can use unit-test availability as a measure

pkg/infradriver/ssh_infradriver.go
```diff
@@ -81,7 +82,8 @@ func NewInfraDriver(cluster *v2.Cluster) (InfraDriver, error) {
 	var err error
 	ret := &SSHInfraDriver{
 		clusterName:       cluster.Name,
-		clusterImagerName: cluster.Spec.Image,
+		clusterImageName:  cluster.Spec.Image,
```
Starnop, line 85: It's confusing here, what's the meaning of clusterImageName and will there be `appImageName` in the future?

pkg/cluster-runtime/installer.go
kakaZhou719: agree with you, application image  should  to be deployed on any master node. here i just chose master0 to execute launch cmds. Do we have some scenarios that require this execute node?

pkg/cluster-runtime/installer.go
kakaZhou719: got it from image extension.Launch.Cmds

pkg/infradriver/ssh_infradriver.go
kakaZhou719: yes, its a string value that just represent the name of this clusterimage to launch.

pkg/cluster-runtime/installer.go
kakaZhou719: the length of masters will not equal to zero during the installation process, that's because we set default master role  in the beginning which was got from user input.

VinceCui (commented):
LGTM

VinceCui (approved):

PR #1765, opened 2022-10-10
Add cmd ut

Stevent-fei:

### DIFF
--- a/cmd/sealer/cmd/utils/cluster_test.go
+++ b/cmd/sealer/cmd/utils/cluster_test.go
@@ -63,11 +63,7 @@ func Test_returnFilteredIPList(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			if res := removeIPList(tt.clusterIPList, tt.toBeDeletedIPList); (res != nil) != tt.wantErr {
-				assert.NotNil(t, res)
-				result := assert.Equal(t, tt.IPListExpected, res)
-				if !result {
-					t.Error("The filtered ip address is different from the expected value")
-				}
+				assert.Equal(t, tt.IPListExpected, res)
 			}
 		})
 	}
--- a/cmd/sealer/cmd/utils/cluster_test.go
+++ b/cmd/sealer/cmd/utils/cluster_test.go
@@ -29,40 +29,35 @@ func Test_returnFilteredIPList(t *testing.T) {
 		clusterIPList     []net.IP
 		toBeDeletedIPList []net.IP
 		IPListExpected    []net.IP
-		wantErr           bool
 	}{
 		{
 			"test",
 			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3"), net.ParseIP("10.10.10.4")},
 			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3"), net.ParseIP("10.10.10.4")},
 			[]net.IP{},
-			false,
 		},
 		{
 			"test1",
 			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3"), net.ParseIP("10.10.10.4")},
 			[]net.IP{},
 			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3"), net.ParseIP("10.10.10.4")},
-			false,
 		},
 		{
 			"test2",
 			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3"), net.ParseIP("10.10.10.4")},
 			[]net.IP{net.ParseIP("10.10.10.4")},
 			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3")},
-			false,
 		},
 		{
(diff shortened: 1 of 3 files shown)


cmd/sealer/cmd/utils/hosts_test.go
```diff
@@ -47,7 +72,26 @@ func Test_getHosts(t *testing.T) {
...
+			var ips net.IP
+			for _, host := range cluster.Spec.Hosts {
+				for _, ips = range host.IPS {
+				}
+			}
+
+			var i net.IP
+			for _, ip := range hosts {
+				for _, i = range ip.IPS {
+				}
```
Starnop, line 89: unfinished code here？

cmd/sealer/cmd/utils/cluster_test.go
```diff
@@ -0,0 +1,203 @@
...
+// limitations under the License.
+
+package utils
+
+import (
+	"fmt"
+	"net"
+	"testing"
+
+	"github.com/sealerio/sealer/cmd/sealer/cmd/types"
```
Starnop, line 22: usually, there should be three groups separated by blank for go-imports

the golang official package
the package for the current project
the others

cmd/sealer/cmd/utils/hosts_test.go
Stevent-fei: The code has been updated

cmd/sealer/cmd/utils/cluster_test.go
```diff
@@ -0,0 +1,201 @@
...
+		{
+			"test3",
+			[]net.IP{},
+			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3"), net.ParseIP("10.10.10.4")},
+			false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if res := removeIPList(tt.clusterIPList, tt.toBeDeletedIPList); (res != nil) != tt.wantErr {
```
justadogistaken, line 62: I don't check the "removeIPList".
But I guess it aims to remove "toBeDeletedIPList" from "clusterIPList", and return the left "ip list"?
I think we could make the test more readable like this:
struct {
		name              string
		IPList     []net.IP
		IPListToDelete []net.IP
                 IPListExpected []net.IP
}

Compare the  "res := removeIPList" and "IPListExpected". Otherwise the test case is hard to be understood.

cmd/sealer/cmd/utils/cluster_test.go
```diff
@@ -15,54 +15,75 @@
...
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			if res := removeIPList(tt.clusterIPList, tt.toBeDeletedIPList); (res != nil) != tt.wantErr {
-				fmt.Println(res)
+				assert.NotNil(t, res)
+				result := assert.Equal(t, tt.IPListExpected, res)
+				if !result {
+					t.Errorf("The filtered ip address is different from the expected value")
```
justadogistaken, line 69: t.Errorf => t.Error

cmd/sealer/cmd/utils/cluster_test.go
```diff
@@ -15,54 +15,75 @@
...
 			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3"), net.ParseIP("10.10.10.4")},
+			[]net.IP{},
 			false,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			if res := removeIPList(tt.clusterIPList, tt.toBeDeletedIPList); (res != nil) != tt.wantErr {
-				fmt.Println(res)
+				assert.NotNil(t, res)
```
justadogistaken, line 66: "(res != nil) != tt.wantErr"
"assert.NotNil(t, res)"
Are them duplicated?
IMO, both of them are unnecessary.
Just do "assert.Equal(t, tt.IPListExpected, res)".
In addition, I guess "if !result " will not be executed if "assert" error happened?

Stevent-fei: Yes, the judgment does repeat itself

cmd/sealer/cmd/utils/cluster_test.go
```diff
@@ -0,0 +1,203 @@
...
+		{
+			"test3",
+			[]net.IP{},
+			[]net.IP{net.ParseIP("10.10.10.1"), net.ParseIP("10.10.10.2"), net.ParseIP("10.10.10.3"), net.ParseIP("10.10.10.4")},
+			[]net.IP{},
+			false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
justadogistaken, line 64: It's not simple enough. Let's finish the last change.
IMO, "(res != nil) != tt.wantErr" means `if res != nil`. In this case, `wantErr` should be removed.
And let's do like this:
1. remove wantErr
2. 
```
res := removeIPList(tt.clusterIPList, tt.toBeDeletedIPList)
if res == nil {
continue
}

assert.Equal(t, tt.IPListExpected, res)
```
Or make the expected IPList to be `nil`
Then you don't have to make sure the `res != nil`.

Stevent-fei: This judgment is inside a for loop, so you can't use continue?



>Or make the expected IPList to be nil

>Then you don't have to make sure the res ! = nil.

The meaning of this sentence is not quite understood

But my understanding is this:

If expected IPList is empty, an error will be reported when the res and expected IPList functions are detected

justadogistaken (approved):
LGTM

PR #1772, opened 2022-10-14
bugfix: support to clean cluster if failed to run cluster image

kakaZhou719:
1. support to clean cluster if failed to run cluster image
2. cluster reset do not need to delete nodes if we delete all cluster.


VinceCui (approved):
LGTM

PR #1774, opened 2022-10-14
support to set host aliases via clusterfile

kakaZhou719:
1. add cluster level host aliases 
2. support to set host aliases at run and scale up
3. support to delete host aliases at delete and scale down


VinceCui (approved):
LGTM

PR #1792, opened 2022-10-21
optimize: load plugins before image distributor

kakaZhou719:

1. support load plugins from cluster image before image distributor.
2. save all mount info, like mount dir, platform, targethosts, after mount cluster image 
3. confirm first when sealer delete


VinceCui (approved):
LGTM

PR #1794, opened 2022-10-22
optimize:no need launch local registry when join node

kakaZhou719:
1. no need launch local registry when join node
2. `InstallOn` will install registry configuration on each given hosts, such as auth info and certs. 
3. `Launch` only start built-in cluster registry component.


VinceCui (approved):
LGTM

PR #1797, opened 2022-10-24
fix: Distribute rootfs

Stevent-fei:
…ory under rootfs

1. Distribute rootfs and ensure that there is an application directory under rootfs


kakaZhou719 (commented):
LGTM

kakaZhou719 (approved):

PR #1799, opened 2022-10-24
support sealer run-app sub command

kakaZhou719:

as we know , we use sealer run to launch cluster image including k8s cluster and build-in apps. meanwhile, sealer also support to build cluster image which image type is application that will only launch build-in apps.

So I suggest that we can separate these two sealer run process and support an additional command.

### DIFF
--- a/cmd/sealer/cmd/cluster/run-app.go
+++ b/cmd/sealer/cmd/cluster/run-app.go
@@ -1,0 +1,144 @@
+// Copyright © 2021 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cluster
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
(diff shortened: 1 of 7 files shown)


cmd/sealer/cmd/cluster/run-app.go
```diff
@@ -0,0 +1,144 @@
...
+			if err != nil {
+				return err
+			}
+
+			extension, err := imageEngine.GetSealerImageExtension(&options.GetImageAnnoOptions{ImageNameOrID: appImageName})
+			if err != nil {
+				return fmt.Errorf("failed to get cluster image extension: %s", err)
+			}
+
+			if extension.Type != v12.AppInstaller {
```
VinceCui, line 86: extension Type should be named as AppImage, rather than AppInstaller

cmd/sealer/cmd/cluster/run-app.go
```diff
@@ -0,0 +1,144 @@
...
+			}()
+
+			distributor, err := imagedistributor.NewScpDistributor(imageMountInfo, infraDriver, nil)
+			if err != nil {
+				return err
+			}
+
+			//todo grab this config from cluster file, that's because it belongs to cluster level information
+			var registryConfig registry.RegConfig
+			var config = registry.Registry{
```
VinceCui, line 119: registry.Registry should be resolved from ClusterFile

kakaZhou719: yes, registry config belongs to cluster level ，i already add todo for those logic.  I'm planning to update the cluster api after this PR merged.

cmd/sealer/cmd/cluster/run-app.go
kakaZhou719: `Type` was already defined at SealerImageExtension. KubeInstaller = "kube-installer" , AppInstaller  = "app-installer"

VinceCui (approved):
LGTM

Issue #1805, opened 2022-10-24
If the current cluster exists, an error should be returned when sealer run


kakaZhou719:
### Other environment you want to tell us?

- Cloud provider or hardware configuration:
- Install tools:
- Others:

PR #1812, opened 2022-10-26
fix: current cluster exists, an error should be returned when sealer run

Stevent-fei:
#1805


kakaZhou719:
only check `common.DefaultKubeConfigFile` is not enough , i think we could try to get nodes to determine whether the cluster is exist.

VinceCui (commented):
When run, should check if k8s is exist.
When apply, no need check

PR #1813, opened 2022-10-26
bugfix: image save use ociv1 format

kakaZhou719:

set default sealer save format as `oci-archive`, thats because we need `ImageAnnotations` to launch cluster image.


pkg/infradriver/ssh_infradriver.go
```diff
@@ -365,19 +365,15 @@ func (d *SSHInfraDriver) Execute(hosts []net.IP, f func(host net.IP) error) erro
 }
 
 func checkAllHostsSameFamily(nodeList []net.IP) error {
```
justadogistaken, line 367: Do we have the unit test for this?

kakaZhou719: oh , i added the UT test.

justadogistaken (approved):

Issue #1802, opened 2022-10-24
If join a node fail, it will be stored in ClusterFile and can't join it again, unless remove it from ClusterFile manually.


VinceCui:
### What happen?

![image](https://user-images.githubusercontent.com/23137196/197527487-90cd788b-c37d-468f-8243-631e1ff2ae05.png)

### Other environment you want to tell us?

- Cloud provider or hardware configuration:
- Install tools:
- Others:

zhy76:
I can help to solve this.

PR #1821, opened 2022-10-27
fix:If join a node fail, it will be stored in ClusterFile and can't join it again

zhy76:
fix:If join a node fail, it will be stored in ClusterFile and can't join it again

fix:If join a node fail, it will be stored in ClusterFile and can't join it again, unless remove it from ClusterFile manually

#1802 
### Describe how you did it
At the beginning of the sealer join cmd,I store the default Clusterfile,if there exits an error,restore the ClusterFile to the default file,the second defer func will infect the first defer func and the value of 'err',so I change it to 'e'.

### Describe how to verify it
I have made some tests,if cmd fail,the Clusterfile will be the default one:
before sealer join:
image
executing sealer join:
image
if sealer join cmd fail,the Clusterfile will be the default:
image


kakaZhou719:
nice, pls sync the same logic to scale-up. thats because join will be replaced by scale-up in the future .

zhy76:
> 



> nice, pls sync the same logic to scale-up. thats because join will be replaced by scale-up in the future .

OK

zhy76:
PTAL

kakaZhou719 (approved):
LGTM

kakaZhou719:
It would be better if you rebase the code.

zhy76:
> It would be better if you rebase the code.
OK

PR #1823, opened 2022-10-27
fix:If join a node fail, it will be stored in ClusterFile and can't join it again

zhy76:
fix:If join a node fail, it will be stored in ClusterFile and can't join it again

fix:If join a node fail, it will be stored in ClusterFile and can't join it again, unless remove it from ClusterFile manually

#1802  
### Describe how you did it
At the beginning of the sealer join cmd,I store the default Clusterfile,if there exits an error,restore the ClusterFile to the default file,the second defer func will infect the first defer func and the value of 'err',so I change it to 'e'.

### Describe how to verify it
I have made some tests,if cmd fail,the Clusterfile will be the default one:
before sealer join:
image
executing sealer join:
image
if sealer join cmd fail,the Clusterfile will be the default:
image

### DIFF
--- a/cmd/sealer/cmd/cluster/join.go
+++ b/cmd/sealer/cmd/cluster/join.go
@@ -76,6 +76,9 @@ func NewJoinCmd() *cobra.Command {
 				return err
 			}
 
+			//store the Cluster as CfSnapshot for rollback
+			cf.CommitSnapshot()
+
 			cluster := cf.GetCluster()
 			if err = utils.ConstructClusterForScaleUp(&cluster, joinFlags, joinMasterIPList, joinNodeIPList); err != nil {
 				return err
@@ -86,6 +89,14 @@ func NewJoinCmd() *cobra.Command {
 			if err = cf.SaveAll(); err != nil {
 				return err
 			}
+
+			defer func() {
+				if err == nil {
+					return
+				}
+				//if there exits an error,rollback the ClusterFile to the default file
+				cf.RollBackClusterFile()
+			}()
 
 			infraDriver, err := infradriver.NewInfraDriver(&cluster)
 			if err != nil {
@@ -117,9 +128,8 @@ func NewJoinCmd() *cobra.Command {
 				return err
 			}
 			defer func() {
-				err = imageMounter.Umount(imageMountInfo)
-				if err != nil {
-					logrus.Errorf("failed to umount cluster image")
+				if e := imageMounter.Umount(imageMountInfo); e != nil {
+					logrus.Errorf("failed to umount cluster image: %v", e)
 				}
 			}()
 
--- a/cmd/sealer/cmd/cluster/scale-up.go
+++ b/cmd/sealer/cmd/cluster/scale-up.go
@@ -76,6 +76,9 @@ func NewScaleUpCmd() *cobra.Command {
 				return err
 			}
 
+			//store the Cluster as CfSnapshot for rollback
+			cf.CommitSnapshot()
+
 			cluster := cf.GetCluster()
 			if err = utils.ConstructClusterForScaleUp(&cluster, ScaleUpFlags, scaleUpMasterIPList, scaleUpNodeIPList); err != nil {
 				return err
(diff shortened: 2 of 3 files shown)


cmd/sealer/cmd/cluster/scale-up.go
```diff
@@ -75,13 +76,27 @@ func NewScaleUpCmd() *cobra.Command {
...
 			if err = utils.ConstructClusterForScaleUp(&cluster, ScaleUpFlags, scaleUpMasterIPList, scaleUpNodeIPList); err != nil {
 				return err
 			}
 			cf.SetCluster(cluster)
 
+			defer func() {
+				if err == nil {
+					return
+				}
+				//if there exits an error,restore the ClusterFile to the default file
```
justadogistaken, line 94: Let's name the function called "rollback or something"? Because I see you call this in two places.

cmd/sealer/cmd/cluster/join.go
```diff
@@ -52,8 +52,9 @@ func NewJoinCmd() *cobra.Command {
 		Example: exampleForJoinCmd,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			var (
-				cf  clusterfile.Interface
-				err error
+				cf        clusterfile.Interface
+				defaultCf clusterfile.Interface
```
justadogistaken, line 56: I guess its functionality is like "cfSnapshot" instead of "default"?

zhy76:
OK,I'll fix it

cmd/sealer/cmd/cluster/join.go
```diff
@@ -76,6 +76,9 @@ func NewJoinCmd() *cobra.Command {
 				return err
 			}
 
+			//store the Cluster as CfSnapshot for rollback
+			cf.CommitCfSnapshot(cf.GetCluster())
```
justadogistaken, line 80: "CommitCfSnapshot" => "CommitSnapshot".
And no need to pass the param "cf.GetCluster()".
Typically, I snapshot things based on current state of object.
Just "CommitCfSnapshot()"

cmd/sealer/cmd/cluster/join.go
```diff
@@ -87,6 +90,13 @@ func NewJoinCmd() *cobra.Command {
 				return err
 			}
 
+			defer func() {
+				//if there exits an error,rollback the ClusterFile to the default file
+				if err = cf.RollBackClusterFile(err); err != nil {
```
justadogistaken, line 95: There is no need to pass error to the function.
This function does not have to be aware of error.
if err == nil {
  return
}

RollBackClusterFile()
...

pkg/clusterfile/clusterfile.go
```diff
@@ -146,6 +153,18 @@ func (c *ClusterFile) SaveAll() error {
 	return utilsos.NewCommonWriter(fileName).WriteFile(bytes.Join(clusterfileBytes, []byte("---\n")))
 }
 
+func (c *ClusterFile) RollBackClusterFile(err error) error {
+	//if there exits an error,rollback the ClusterFile to the default file
+	if err == nil {
+		return nil
+	}
+	c.cluster = c.cfSnapshot
```
justadogistaken, line 161: This makes "cluster" and "cfSnapshot" refer to a same address.
What if we don't exit process after RollBackClusterFile()?
After "RollBackClusterFile", what we do to "cluster" will also impact "cfSnapshot".
We'd better clone "cfSnapshot", and give it to "cluster".
And check "cfSnapshot == nil"

zhy76: Thanks,I'll optimsize them.

pkg/clusterfile/clusterfile.go
```diff
@@ -146,6 +155,15 @@ func (c *ClusterFile) SaveAll() error {
 	return utilsos.NewCommonWriter(fileName).WriteFile(bytes.Join(clusterfileBytes, []byte("---\n")))
 }
 
+func (c *ClusterFile) RollBackClusterFile() {
```
justadogistaken, line 158: throw a error if "cfSnapshot" == nil.

justadogistaken:
@kakaZhou719 @VinceCui @Stevent-fei PTAL

kakaZhou719 (approved):
LGTM

PR #1825, opened 2022-10-28
fix:sealer run bug

Stevent-fei:

Fix the sealer run command failure
image


kakaZhou719 (approved):
LGTM

PR #1850, opened 2022-11-11
sealer run-app supports concurrent  run 

kakaZhou719:

At present, our cluster image data is mixed with application image data and base image data, If we run multiple images at the same time, there will be conflicts.

1. Detach the mount directory
all imagedata mount directory will under below directory:
`/var/lib/sealer/data/mount/${imageName}/${platform}`

all application imagedata rootfs directory will under below directory:
`var/lib/sealer/data/${imageName}/rootfs/application/apps`

all basefs  imagedata directory will under below directory:
`var/lib/sealer/data/${imageName}/rootfs`

2. add a file lock on .sealer/application.json

### DIFF
--- a/pkg/cluster-runtime/apps.go
+++ b/pkg/cluster-runtime/apps.go
@@ -15,12 +15,14 @@
 package clusterruntime
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net"
+	"os"
 	"path/filepath"
+	"syscall"
+
+	"github.com/sirupsen/logrus"
 
 	"github.com/moby/buildkit/frontend/dockerfile/shell"
 	"github.com/sealerio/sealer/common"
@@ -29,7 +31,6 @@ import (
 	"github.com/sealerio/sealer/pkg/imagedistributor"
 	"github.com/sealerio/sealer/pkg/infradriver"
 	"github.com/sealerio/sealer/pkg/registry"
-	osutils "github.com/sealerio/sealer/utils/os"
 )
 
 type AppInstaller struct {
@@ -50,7 +51,7 @@ func NewAppInstaller(infraDriver infradriver.InfraDriver, distributor imagedistr
 
 func (i *AppInstaller) Install(master0 net.IP, cmds []string) error {
 	// distribute rootfs
-	if err := i.distributor.DistributeRootfs([]net.IP{master0}, i.infraDriver.GetClusterRootfsPath()); err != nil {
+	if err := i.distributor.Distribute([]net.IP{master0}, i.infraDriver.GetClusterRootfsPath()); err != nil {
 		return err
 	}
 
@@ -78,9 +79,9 @@ func (i *AppInstaller) Install(master0 net.IP, cmds []string) error {
 
 func (i AppInstaller) LaunchClusterImage(master0 net.IP, launchCmds []string) error {
 	var (
-		cmds          []string
-		clusterRootfs = i.infraDriver.GetClusterRootfsPath()
-		ex            = shell.NewLex('\\')
+		cmds    []string
+		appPath = i.infraDriver.GetClusterRootfsPath()
+		ex      = shell.NewLex('\\')
 	)
 
 	if len(launchCmds) > 0 {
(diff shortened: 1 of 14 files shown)


pkg/cluster-runtime/apps.go
```diff
@@ -78,9 +76,9 @@ func (i *AppInstaller) Install(master0 net.IP, cmds []string) error {
 
 func (i AppInstaller) LaunchClusterImage(master0 net.IP, launchCmds []string) error {
 	var (
-		cmds          []string
-		clusterRootfs = i.infraDriver.GetClusterRootfsPath()
-		ex            = shell.NewLex('\\')
+		cmds    []string
+		appPath = i.infraDriver.GetClusterRootfsPath()
```
Starnop, line 80: It looks like we can keep use `clusterRootfs`.

And In my understanding, we may need a new `appRootPath` variable, and it's going to be filepath.Join(clusterRootfs,"applications")

pkg/imagedistributor/mount.go
```diff
@@ -19,22 +19,26 @@ import (
...
-	"github.com/sealerio/sealer/common"
 	imagecommon "github.com/sealerio/sealer/pkg/define/options"
 	"github.com/sealerio/sealer/pkg/imageengine"
 	v1 "github.com/sealerio/sealer/types/api/v1"
 	osi "github.com/sealerio/sealer/utils/os"
 	"github.com/sealerio/sealer/utils/os/fs"
 )
 
+const (
+	mountBaseDir = "/var/lib/sealer/data/mount"
```
Starnop, line 32: move it to. common/common.go file?

pkg/imagedistributor/mount.go
```diff
@@ -50,7 +54,13 @@ func (b buildAhMounter) Mount(imageName string, platform v1.Platform) (string, e
 		return "", err
 	}
 
-	if _, err := b.imageEngine.CreateWorkingContainer(&imagecommon.BuildRootfsOptions{
+	// make sure base mount Dir is existed.
+	err := fs.FS.MkdirAll(filepath.Dir(mountDir))
+	if err != nil {
```
Starnop, line 59: consider the following syntax to avoid widening the scope of error

if err := fs.FS.MkdirAll(filepath.Dir(mountDir));err!=nil{}

pkg/imagedistributor/mount.go
```diff
@@ -108,13 +118,17 @@ func (c ImagerMounter) Mount(imageName string) ([]ClusterImageMountInfo, error)
...
-func (c ImagerMounter) Umount(imageMountInfo []ClusterImageMountInfo) error {
+func (c ImagerMounter) Umount(imageName string, imageMountInfo []ClusterImageMountInfo) error {
 	for _, info := range imageMountInfo {
 		err := c.Mounter.Umount(info.MountDir)
 		if err != nil {
 			return fmt.Errorf("failed to umount %s:%v", info.MountDir, err)
 		}
 	}
+	// delete all mounted images
+	if err := fs.FS.RemoveAll(filepath.Join(mountBaseDir, formatImageName(imageName))); err != nil {
```
Starnop, line 129: how about extracting a public method to return the mount root directory for the specified image name?

pkg/imagedistributor/mount.go
```diff
@@ -125,3 +139,7 @@ func NewImageMounter(imageEngine imageengine.Interface, hostsPlatform map[v1.Pla
 	c.Mounter = NewBuildAhMounter(imageEngine)
 	return c, nil
 }
+
+func formatImageName(name string) string {
```
Starnop, line 143: add some unit tests and try to consider more scenarios

pkg/infradriver/ssh_infradriver.go
```diff
@@ -337,11 +338,11 @@ func (d *SSHInfraDriver) GetHostsPlatform(hosts []net.IP) (map[v1.Platform][]net
 }
 
 func (d *SSHInfraDriver) GetClusterRootfsPath() string {
-	return common.DefaultTheClusterRootfsDir(d.clusterName)
+	return filepath.Join(common.DefaultSealerDataDir, d.clusterName, "rootfs")
 }
 
 func (d *SSHInfraDriver) GetClusterBasePath() string {
-	return common.DefaultClusterBaseDir(d.clusterName)
+	return filepath.Join(common.DefaultSealerDataDir, d.clusterName)
```
Starnop, line 345: why not keep it?

pkg/infradriver/ssh_infradriver.go
```diff
@@ -337,11 +338,11 @@ func (d *SSHInfraDriver) GetHostsPlatform(hosts []net.IP) (map[v1.Platform][]net
 }
 
 func (d *SSHInfraDriver) GetClusterRootfsPath() string {
-	return common.DefaultTheClusterRootfsDir(d.clusterName)
+	return filepath.Join(common.DefaultSealerDataDir, d.clusterName, "rootfs")
```
Starnop, line 341: why not keep it?

pkg/cluster-runtime/apps.go
```diff
@@ -98,38 +96,39 @@ func (i AppInstaller) LaunchClusterImage(master0 net.IP, launchCmds []string) er
...
+	return i.save(common.GetDefaultApplicationFile())
 }
 
 // todo save image info to disk or api server, we need new interface to do this.
-func (i AppInstaller) save() error {
-	var extensionList []v12.ImageExtension
-	applicationFile := common.GetDefaultApplicationFile()
+func (i AppInstaller) save(applicationFile string) error {
+	f, err := os.OpenFile(applicationFile, os.O_RDWR|os.O_APPEND|os.O_CREATE, 0600)
+	defer func() {
```
Starnop, line 110: use `defer f.Close() ` is OK

pkg/cluster-runtime/apps.go
```diff
@@ -98,38 +96,39 @@ func (i AppInstaller) LaunchClusterImage(master0 net.IP, launchCmds []string) er
...
-			}
-		}
+	err = syscall.Flock(int(f.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)
+	if err != nil {
+		return fmt.Errorf("cannot flock file %s - %s", applicationFile, err)
 	}
 
-	extensionList = append(extensionList, i.extension)
-	content, err := json.Marshal(extensionList)
+	content, err := json.Marshal(i.extension)
```
Starnop, line 123: user defer func to unlock file locker here

pkg/cluster-runtime/apps.go
```diff
@@ -98,38 +96,39 @@ func (i AppInstaller) LaunchClusterImage(master0 net.IP, launchCmds []string) er
...
 	if err != nil {
 		return fmt.Errorf("failed to marshal image extension: %v", err)
 	}
 
-	return osutils.NewCommonWriter(applicationFile).WriteFile(content)
+	_, err = f.Write(content)
+	if err != nil {
```
Starnop, line 129: unify use 
if err:=func();err!=nil{}

please check the whole picture for yourself

pkg/cluster-runtime/apps.go
```diff
@@ -98,38 +96,39 @@ func (i AppInstaller) LaunchClusterImage(master0 net.IP, launchCmds []string) er
...
 
-	extensionList = append(extensionList, i.extension)
-	content, err := json.Marshal(extensionList)
+	content, err := json.Marshal(i.extension)
 	if err != nil {
 		return fmt.Errorf("failed to marshal image extension: %v", err)
 	}
 
-	return osutils.NewCommonWriter(applicationFile).WriteFile(content)
+	_, err = f.Write(content)
```
Starnop, line 128: I find these codes changed the existing logic and the old logic will append the extensions from the old applicationFile.

Cloud you please make a simple explanation?

pkg/cluster-runtime/apps.go
kakaZhou719: yes, we should launch app in their own workspace, thats why i rename this value from "clusterRootfs" to "appRootPath". later we only need to update this func to support it directly.

pkg/imagedistributor/mount.go
kakaZhou719: mountBaseDir is internal used , so i dont move it to pubic common.

pkg/infradriver/ssh_infradriver.go
kakaZhou719: here i thought, the only way we get the cluster value such as "GetClusterRootfsPath" and the same kind parameters. not from common pkg directly .

pkg/cluster-runtime/apps.go
kakaZhou719: golang lint will check the func return error ,to make sure each error was handled . so i use defer func()

pkg/cluster-runtime/apps.go
kakaZhou719: Currently ,it's also record app installations with append option, just separated from each installation record.

pkg/infradriver/ssh_infradriver.go
Starnop: And then, should we delete the function `common.DefaultTheClusterRootfsDir`?

Starnop (approved):

PR #1851, opened 2022-11-11
saving container image optional

### LINKED ISSUE #1849: sealer build make saving container image feature is optional
## Issue Description

Type: *feature request*

in some case, users want to not save the container image generated during the build process . if so,  the size of the application image will be much smaller.

we can add a flag `--download-container-image`, Or `--lite` , and its default value is true.

YTGhost:

Fixes #1849

### DIFF
--- a/cmd/sealer/cmd/image/build.go
+++ b/cmd/sealer/cmd/image/build.go
@@ -92,6 +92,7 @@ func NewBuildCmd() *cobra.Command {
 	buildCmd.Flags().StringSliceVar(&buildFlags.Annotations, "annotation", []string{}, "add annotations for image. Format like --annotation key=[value]")
 	buildCmd.Flags().StringSliceVar(&buildFlags.Labels, "label", []string{getSealerLabel()}, "add labels for image. Format like --label key=[value]")
 	buildCmd.Flags().BoolVar(&buildFlags.NoCache, "no-cache", false, "do not use existing cached images for building. Build from the start with a new set of cached layers.")
+	buildCmd.Flags().BoolVarP(&buildFlags.DownloadContainerImage, "download-container-image", "d", true, "save the container image generated during the build process.")
 
 	supportedImageType := map[string]struct{}{v12.KubeInstaller: {}, v12.AppInstaller: {}}
 	if _, ok := supportedImageType[buildFlags.ImageType]; !ok {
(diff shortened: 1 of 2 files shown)


cmd/sealer/cmd/image/build.go
```diff
@@ -160,13 +161,15 @@ func buildSealerImage() error {
...
 
 	defer func() {
-		for _, m := range []string{tempTag} {
-			// the above image is intermediate image, we need to remove it when the build ends.
-			if err := engine.RemoveImage(&bc.RemoveImageOptions{
-				ImageNamesOrIDs: []string{m},
-				Force:           true,
-			}); err != nil {
-				logrus.Debugf("failed to remove image %s, you need to remove it manually: %v", m, err)
+		if buildFlags.DownloadContainerImage {
```
kakaZhou719, line 164: we could skip download the container image if it is false, we do not need to do `applyRegistryToImage`

cmd/sealer/cmd/image/build.go
```diff
@@ -92,6 +92,7 @@ func NewBuildCmd() *cobra.Command {
 	buildCmd.Flags().StringSliceVar(&buildFlags.Annotations, "annotation", []string{}, "add annotations for image. Format like --annotation key=[value]")
 	buildCmd.Flags().StringSliceVar(&buildFlags.Labels, "label", []string{getSealerLabel()}, "add labels for image. Format like --label key=[value]")
 	buildCmd.Flags().BoolVar(&buildFlags.NoCache, "no-cache", false, "do not use existing cached images for building. Build from the start with a new set of cached layers.")
+	buildCmd.Flags().BoolVarP(&buildFlags.DownloadContainerImage, "download-container-image", "lite", true, "save the container image generated during the build process.")
```
kakaZhou719, line 95: we use "download-container-image" Or "lite" to set this flag`s name , and for flag`s shorthand, we usually keep the initial letter of the word.

cmd/sealer/cmd/image/build.go
```diff
@@ -92,6 +92,7 @@ func NewBuildCmd() *cobra.Command {
 	buildCmd.Flags().StringSliceVar(&buildFlags.Annotations, "annotation", []string{}, "add annotations for image. Format like --annotation key=[value]")
 	buildCmd.Flags().StringSliceVar(&buildFlags.Labels, "label", []string{getSealerLabel()}, "add labels for image. Format like --label key=[value]")
 	buildCmd.Flags().BoolVar(&buildFlags.NoCache, "no-cache", false, "do not use existing cached images for building. Build from the start with a new set of cached layers.")
+	buildCmd.Flags().BoolVarP(&buildFlags.DownloadContainerImage, "download-container-image", "d", true, "save the container image generated during the build process.")
```
Starnop, line 95: how about adding a new flag like `buildMode`?

And different build mode means difference logic of execution

PR #1854, opened 2022-11-11
fix plugin bug

Stevent-fei:
When running the plugin script, the values in the plugins do not take effect.
I have found the cause of the problem: there is a problem with the repeated judgment function in plugins, which causes the plugin to not execute.
so i fix it.

### DIFF
--- a/pkg/plugin/plugins.go
+++ b/pkg/plugin/plugins.go
@@ -23,15 +23,13 @@ import (
 	"plugin"
 	"strings"
 
-	"github.com/sealerio/sealer/utils"
-
-	"github.com/sealerio/sealer/utils/yaml"
-
 	"github.com/sealerio/sealer/common"
 	v1 "github.com/sealerio/sealer/types/api/v1"
 	v2 "github.com/sealerio/sealer/types/api/v2"
+	"github.com/sealerio/sealer/utils"
 	"github.com/sealerio/sealer/utils/platform"
 	strUtils "github.com/sealerio/sealer/utils/strings"
+	"github.com/sealerio/sealer/utils/yaml"
 )
 
 type InvalidPluginTypeError struct {
@@ -74,6 +72,7 @@ func (c *PluginsProcessor) Load() error {
 	if err != nil {
 		return fmt.Errorf("failed to load plugin dir: %v", err)
 	}
+
 	for _, f := range files {
 		// load shared object(.so) file
 		if filepath.Ext(f.Name()) == ".so" {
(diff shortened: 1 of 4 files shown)


kakaZhou719:
@Stevent-fei ，could you pls describe this bug more specifically which is  helpful for others to read it.

pkg/plugin/plugins.go
```diff
@@ -74,30 +74,26 @@ func (c *PluginsProcessor) Load() error {
...
-				return err
-			}
-			Register(pt, p)
-		}
-		if yaml.Matcher(f.Name()) {
-			plugins, err := utils.DecodeCRDFromFile(filepath.Join(path, f.Name()), common.Plugin)
-			if err != nil {
-				return fmt.Errorf("failed to load plugin: %v", err)
+
+	if files != nil {
```
Starnop, line 78: if the files ==nil, 

the logic like `for _, f := range files` will not be executed actually

So no need to change that, just use `for _, f := range files` is better

pkg/plugin/plugins.go
```diff
@@ -74,30 +74,26 @@ func (c *PluginsProcessor) Load() error {
...
+		c.Plugins = nil
```
Starnop, line 79: why reset c.Plugins here?

Starnop:
image

It is better to use `git commit-s --amend` or `git rebase with squash`

pkg/plugin/plugins.go
```diff
@@ -87,23 +86,36 @@ func (c *PluginsProcessor) Load() error {
...
+			c.Plugins = append(c.Plugins, plugins.([]v1.Plugin)...)
+			i := deduplication(c.Plugins)
+			c.Plugins = i
 		}
 	}
 
 	return nil
 }
 
+func deduplication(plugins []v1.Plugin) []v1.Plugin {
```
Starnop, line 101: so what has changed

Stevent-fei: The previous function did not do all the deduplication

27149chen:
what is the status of this pr?

Stevent-fei:
> 

@27149chen The plugin bug has been fixed.

27149chen:
> > 
> 
> @27149chen The plugin bug has been fixed.

@Stevent-fei in which version? I didn't find any newer releases?

Stevent-fei:
> > > 
> > 
> > 
> > @27149chen The plugin bug has been fixed.
> 
> @Stevent-fei in which version? I didn't find any newer releases.

@27149chen this change has not been merged, you can use this branch to build new binaries

27149chen:
> > > > 
> > > 
> > > 
> > > @27149chen The plugin bug has been fixed.
> > 
> > 
> > @Stevent-fei in which version? I didn't find any newer releases.
> 
> @27149chen this change has not been merged, you can use this branch to build new binaries

sorry, I can not use a private branch which is not reviewed or approved by anyone

kakaZhou719 (approved):
LGTM

PR #1857, opened 2022-11-12
support only load image apply mode

Stevent-fei:
In special scenarios, we want to load images to the specified registry in advance.

This PR try to support the above feature by do these things:

add apply mode flag
support only loadImage apply mode

### DIFF
--- a/cmd/sealer/cmd/cluster/apply.go
+++ b/cmd/sealer/cmd/cluster/apply.go
@@ -39,6 +39,8 @@ import (
 
 var applyClusterFile string
 
+var applyMode string
+
 const MasterRoleLabel = "node-role.kubernetes.io/master"
 
 var longApplyCmdDescription = `apply command is used to apply a Kubernetes cluster via specified Clusterfile.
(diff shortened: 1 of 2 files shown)


cmd/sealer/cmd/cluster/apply.go
```diff
@@ -142,67 +145,103 @@ func createNewCluster(clusterImageName string, infraDriver infradriver.InfraDriv
...
+		distributor, err := imagedistributor.NewScpDistributor(imageMountInfo, infraDriver, cf.GetConfigs())
+		if err != nil {
+			return err
+		}
 
-	plugins, err := loadPluginsFromImage(imageMountInfo)
-	if err != nil {
-		return err
-	}
+		if err := distributor.DistributeRegistry(infraDriver.GetHostIPList()[0], infraDriver.GetClusterRootfsPath()); err != nil {
```
Starnop, line 171: if applyMode == common.ApplyModeLoadImage {
if err := distributor.DistributeRegistry(infraDriver.GetHostIPList()[0], infraDriver.GetClusterRootfsPath());err!=nil{}

return nil 

}

Starnop: Is it possible to implement the loadImage apply mode by the above few lines of code?

Starnop (approved):

PR #1864, opened 2022-11-15
add --image-list imageList

Stevent-fei:
add imageList Flag to sealer build to support download extra container image rather than copy it to rootfs,Provide additional image download method.
eg:
sealer build -t abc:v1 --image-list imageList

### DIFF
--- a/cmd/sealer/cmd/image/build.go
+++ b/cmd/sealer/cmd/image/build.go
@@ -27,13 +27,16 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sealerio/sealer/build/buildimage"
 	"github.com/sealerio/sealer/build/kubefile/parser"
+	"github.com/sealerio/sealer/common"
 	version2 "github.com/sealerio/sealer/pkg/define/application/version"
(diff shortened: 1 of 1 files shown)


kakaZhou719:
@Stevent-fei ，i am ok with this logic, while i thought could we unify the download container image logic here. for example : grab all images list we needed via walking build context, and then download everyone.

cmd/sealer/cmd/image/build.go
```diff
@@ -253,19 +256,37 @@ func applyRegistryToImage(imageID, tag, manifest string, platform v1.Platform, e
 		return errors.Wrapf(err, "failed to create working container, imageid: %s", imageID)
 	}
 
+	// download container image form `imageList`
```
Starnop, line 259: merge with code 

```
if err := buildimage.NewMiddlewarePuller(v1.Platform{
		Architecture: arch,
		OS:           _os,
		Variant:      variant,
......
```

VinceCui (approved):
LGTM

PR #1873, opened 2022-11-17
feat: save clusterfile to the cluster 

kakaZhou719:

at present , we store the cluster configuration information on the master0 local disk after we create the cluster, such as node auth info , cluster plugins, cluster env and so on .

If this file is mistakenly deleted or modified by mistake, it will be a great disaster for the maintenance of the cluster. 

so we need to store this file in the cluster to do disaster preparedness, and each time we can prioritized use the clusterfile  in the cluster for operation.

### Describe how you did it

1. save clusterfile to cluster as configmap with name "sealer-clusterfile"  in namespace "kub-system".
2. modify some cmd examples

### DIFF
--- a/cmd/sealer/cmd/cluster/apply.go
+++ b/cmd/sealer/cmd/cluster/apply.go
@@ -23,11 +23,9 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sealerio/sealer/common"
 	"github.com/sealerio/sealer/pkg/client/k8s"
-	clusterruntime "github.com/sealerio/sealer/pkg/cluster-runtime"
 	"github.com/sealerio/sealer/pkg/clusterfile"
 	v12 "github.com/sealerio/sealer/pkg/define/image/v1"
 	"github.com/sealerio/sealer/pkg/define/options"
-	"github.com/sealerio/sealer/pkg/imagedistributor"
 	"github.com/sealerio/sealer/pkg/imageengine"
 	"github.com/sealerio/sealer/pkg/infradriver"
 	v2 "github.com/sealerio/sealer/types/api/v2"
@@ -80,10 +78,6 @@ func NewApplyCmd() *cobra.Command {
 				return err
 			}
 
-			//save desired clusterfile
-			if err = cf.SaveAll(); err != nil {
-				return err
-			}
 			desiredCluster := cf.GetCluster()
 			infraDriver, err := infradriver.NewInfraDriver(&desiredCluster)
 			if err != nil {
@@ -98,13 +92,6 @@ func NewApplyCmd() *cobra.Command {
 				return err
 			}
 
-			client := getClusterClient()
-			if client == nil {
-				// no k8s client means to init a new cluster.
-				logrus.Infof("start to create new cluster with image: %s", imageName)
-				return createNewCluster(imageName, infraDriver, imageEngine, cf)
-			}
-
 			extension, err := imageEngine.GetSealerImageExtension(&options.GetImageAnnoOptions{ImageNameOrID: imageName})
 			if err != nil {
 				return fmt.Errorf("failed to get cluster image extension: %s", err)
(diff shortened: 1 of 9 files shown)


pkg/clusterfile/clusterfile.go
```diff
@@ -150,27 +176,72 @@ func (c *ClusterFile) SaveAll() error {
...
+	configMap.Data = map[string]string{configMapDataName: string(data)}
+
+	_, err = client.ConfigMap(configMapNamespace).Apply(context.TODO(), configMap, metav1.ApplyOptions{FieldManager: "kubectl-create"})
+	if err != nil {
+		return fmt.Errorf("failed to create configmap: %v", err)
 	}
+	return nil
 }
 
 func NewClusterFile(b []byte) (Interface, error) {
```
VinceCui, line 208: good design~

pkg/clusterfile/clusterfile.go
```diff
@@ -74,7 +89,18 @@ func (c *ClusterFile) CommitSnapshot() {
 	*c.cfSnapshot = *c.cluster
```
VinceCui, line 89: Although this is not part of this PR, i want to say the cfSnapshot maybe a bad design.

Why we need a snapshot of a cluster file? I think the only trusty cluster status is not the clusterfile but the k8s, so i think we no need control the version of clusterfile carefully.

pkg/clusterfile/clusterfile.go
```diff
@@ -30,17 +34,28 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	configMapNamespace = "kube-system"
+	configMapDataName  = "Clusterfile"
+	configMapName      = "sealer-clusterfile"
+)
+
 type Interface interface {
```
VinceCui, line 43: Shall we save the application.json into k8s too?

cmd/sealer/cmd/cluster/delete.go
```diff
@@ -96,18 +93,28 @@ func NewDeleteCmd() *cobra.Command {
 }
 
 func deleteCluster(workClusterfile string) error {
```
VinceCui, line 95: good design.

cmd/sealer/cmd/cluster/scale-up.go
```diff
@@ -39,8 +38,8 @@ User can scale-up cluster by explicitly specifying host IP`
 
 var exampleForScaleUpCmd = `
 scale-up cluster:
-  sealer join --masters 192.168.0.1 --nodes 192.168.0.2 -p Sealer123
-  sealer join --masters 192.168.0.1-192.168.0.3 --nodes 192.168.0.4-192.168.0.6 -p Sealer123
+  sealer scale-up --masters 192.168.0.1 --nodes 192.168.0.2 -p Sealer123
```
VinceCui, line 41: Are you find that, when you change the join.go or apply.go, you need change this file too. This is a bad design, you should use a common function for both command join, scale-up and apply. 

When should we refactor the code? The answer is NOW.

cmd/sealer/cmd/cluster/apply.go
```diff
@@ -81,7 +81,7 @@ func NewApplyCmd() *cobra.Command {
 			}
 
 			//save desired clusterfile
-			if err = cf.SaveAll(); err != nil {
```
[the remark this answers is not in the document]
VinceCui: If this clusterfile is an App File, should we save it?

VinceCui:
Maybe we can provide SDK for read the ClusterFile from k8s.

pkg/clusterfile/clusterfile.go
kakaZhou719: i will do it in the another PR.

cmd/sealer/cmd/cluster/apply.go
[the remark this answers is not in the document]
kakaZhou719: Mainly to store the expected cluster node information temporarily.  if apply failed, will use it to delete the cluster node ,if not, new save all will overwrite it.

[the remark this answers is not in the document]
kakaZhou719: > If this clusterfile is an App File, should we save it?

yes, if it is app, no need to pre save expected clusterfile.

VinceCui (approved):

PR #1895, opened 2022-11-28
doc: add master0 recovery design

kakaZhou719:

Master0 plays an important role in sealer cluster, it hosts many important components such as sealer builtin registry,
application images launched history and others cluster configuration files. if the master0 node is abnormal, a highly
available solution is required to recover on other master nodes.

Currently, we have plans to save clusterfile to the cluster. while for built registry ,it is still deployed a single
instances on master0 node. So the following document are mainly for built registry.

### DIFF
--- a/docs/design/master0-recovery.md
+++ b/docs/design/master0-recovery.md
@@ -1,0 +1,111 @@
+# Master0 recovery design
+
+## Motivations
+
+Master0 plays an important role in sealer cluster, it hosts many important components such as sealer builtin registry,
+application images launched history and others cluster configuration files. if the master0 node is abnormal, a highly
+available solution is required to recover it on other master nodes.
+
+Currently, we have plans to save clusterfile to the cluster. while for built registry ,it is still deployed a single
+instance on master0 node. So the following document are mainly for built registry.
+
+## Proposal
+
+### For builtin registry
+
+* expose registry configuration to user thought clusterfile.
+* distribute all registry data split from cluster image registry directory to each deploy host, whether it is a cluster
+  image or an application image.
+* save all registry configuration generated form sealer such as, basic auth information and https certificate.
+* save the boot history of each application image information such as tag,version and image id to cluster, which was
+  launched through the sealer run-app.
+
+## Implementation
+
+### For configuration of builtin registry
+
+1. expose below `Registry` struct to clusterfile which is contains configurations about local registry and remote
+   registry.
+
+```yaml
+type Registry struct {
+  Domain        string        `json:"domain,omitempty"`
+  Port          int           `json:"port,omitempty"`
+  Username      string        `json:"username,omitempty"`
+  Password      string        `json:"password,omitempty"`
+  LocalRegistry LocalRegistry `json:"local_registry,omitempty"`
+}
+  type LocalRegistry struct {
+  // DeployHosts is the target host list that local registry will be deployed on.
+  // if not set ,master0 will be the default value.
+  DeployHosts []net.IP `json:"deploy_hosts,omitempty"`
+  // InsecureMode indicated that whether the local registry is exposed in HTTPS.
+  // if true sealer will generate default ssl cert.
+  InsecureMode bool `json:"insecure_mode,omitempty,omitempty"`
+}
+```
+
+2. save registry configuration to cluster as configmap with name "sealer-registry" in namespace "kub-system".
+
+```shell
+# must exist, this is core registry configuration
+/var/lib/sealer/data/my-cluster/rootfs/etc/registry.yml
+#if local registry is not booted on insecure mode
+/var/lib/sealer/data/my-cluster/rootfs/certs/sea.hub.crt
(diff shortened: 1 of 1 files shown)


docs/design/master0-recovery.md
```diff
@@ -0,0 +1,111 @@
...
+  // DeployHosts is the target host list that local registry will be deployed on.
+  // if not set ,master0 will be the default value.
+  DeployHosts []net.IP `json:"deploy_hosts,omitempty"`
+  // InsecureMode indicated that whether the local registry is exposed in HTTPS.
+  // if true sealer will generate default ssl cert.
+  InsecureMode bool `json:"insecure_mode,omitempty,omitempty"`
+}
+```
+
+2. save registry configuration to cluster as configmap with name "sealer-registry" in namespace "kub-system".
```
VinceCui, line 48: use secret not configmap.

kube-system

docs/design/master0-recovery.md
```diff
@@ -0,0 +1,111 @@
...
+* load registry configuration from cluster on new master0
+* dump registry configuration files from cluster to rootfs on new master0
+* distribute all registry data to recovery host thought boot history loaded from cluster on new master0
+* launch builtin registry thought rootfs scripts and configuration files to recover registry service.
+* replace the latest registry ip address on all cluster nodes.
+
+3. also need a sealer subcommand to do above recovery implementation
+
+```shell
+sealer recover registry --host 192.168.1.100 --password abctest
```
VinceCui, line 110: maybe sealer recover master0

docs/design/master0-recovery.md
```diff
@@ -0,0 +1,111 @@
...
+* dump registry configuration files from cluster to rootfs on new master0
+* launch builtin registry thought rootfs scripts and configuration files to recover registry service.
+* replace the latest registry ip address on all cluster nodes.
+
+2. Post-distribute registry data on recovery host
+
+* load registry configuration from cluster on new master0
+* dump registry configuration files from cluster to rootfs on new master0
+* distribute all registry data to recovery host thought boot history loaded from cluster on new master0
+* launch builtin registry thought rootfs scripts and configuration files to recover registry service.
```
VinceCui, line 104: I prefer to distribute registry data to each master, so that each master is equal. If use deployhost, do users have to manually handle the distributing the registry to another new deployhost after recovering once?

VinceCui (commented):
Good design, please see my comments.

docs/design/master0-recovery.md
kakaZhou719: we can also save to others master`s local disk at the same time

docs/design/master0-recovery.md
kakaZhou719: i agree with the point that make all masters is consistent,  `post-distribute` is more flexibility than `pre-distribute` , Both needs to manually  specify on which node the registry service is recover to. The difference is that `pre-distribute` can save the original same data with master0, and the `post-distribute` save the later recovered data according to the history.

docs/design/master0-recovery.md
VinceCui: no need do this. k8s use etcd to provide HA storage.

docs/design/master0-recovery.md
VinceCui: we can implemet pre-distribute firstly.

docs/design/master0-recovery.md
```diff
@@ -0,0 +1,111 @@
...
+  Password      string        `json:"password,omitempty"`
+  LocalRegistry LocalRegistry `json:"local_registry,omitempty"`
+}
+  type LocalRegistry struct {
+  // DeployHosts is the target host list that local registry will be deployed on.
+  // if not set ,master0 will be the default value.
+  DeployHosts []net.IP `json:"deploy_hosts,omitempty"`
+  // InsecureMode indicated that whether the local registry is exposed in HTTPS.
+  // if true sealer will generate default ssl cert.
+  InsecureMode bool `json:"insecure_mode,omitempty,omitempty"`
```
Starnop, line 44: Insecure is OK

docs/design/master0-recovery.md
Starnop, line 104: > launch builtin registry thought rootfs scripts and configuration files to recover registry service.

How about starting a registry by default for each master node?

docs/design/master0-recovery.md
Starnop: How about `sealer sync registry --host 192.168.1.100 --password abctest`？

With that, we can easily sync the registry data to the newly joined master node. Probably it‘s a more general approach in my opinion.

VinceCui (approved):
LGTM

docs/design/master0-recovery.md
kakaZhou719: good idea !

docs/design/master0-recovery.md
kakaZhou719: "InsecureMode" is a noun word ,I  thought  it's better to use this.

PR #1902, opened 2022-12-02
move search to alpha

Stevent-fei:
Support search to view the current list of supported images
1. move search to alpha
2. modify default defaultDomain and defaultRepo


kakaZhou719 (approved):
LGTM

PR #1909, opened 2022-12-07
New design for basefs.

VinceCui:

### DIFF
--- a/docs/api/cloudrootfs.md
+++ b/docs/api/cloudrootfs.md
@@ -4,7 +4,10 @@ cloud rootfs will package all the dependencies refers to the kubernetes cluster 
 
 ```shell script
 .
-├── bin
+├── Metadata # Metadata will be see in the `sealer inspect`
+├── README.md # Metadata will be see in the `sealer help`
+├── imageList # image list in it will be saved into the ClusterImage
+├── bin # binaries will be installed at all nodes' /usr/local/bin
 │   ├── conntrack
 │   ├── containerd-rootless-setuptool.sh
 │   ├── containerd-rootless.sh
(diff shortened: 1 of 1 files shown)


docs/api/cloudrootfs.md
```diff
@@ -13,44 +13,28 @@ cloud rootfs will package all the dependencies refers to the kubernetes cluster
...
-│   ├── docker
-│   ├── dockerd
-│   ├── docker-init
-│   ├── docker-proxy
-│   ├── rootlesskit
-│   ├── rootlesskit-docker-proxy
-│   ├── runc
-│   └── vpnkit
-├── etc
+├── etc # configs will be put into all nodes' sealer-rootfs
```
allencloud, line 19: What does sealer-rootfs?  Could you help to define that? Since there are quite a lot of rootfs in sealer's concept. @VinceCui

docs/api/cloudrootfs.md
```diff
@@ -4,7 +4,7 @@ cloud rootfs will package all the dependencies refers to the kubernetes cluster
 
 ```shell script
 .
-├── bin
+├── bin # binaries will be installed at all nodes' /usr/local/bin
```
allencloud, line 7: If we only store binaries in this directory, I am wondering  whether it is proper to container scripts here, like  `containerd-rootless-setuptool.sh`, `containerd-rootless.sh` and `kubelet-pre-start.sh`.

In addition, binary may not be compatible with plarforms(x86, arm). Then in the design, we make a ClusterImage to be compatible to all arch platforms or different ClusterImage compatible to different platforms. 

It could be better that we have a top-down through design, I think. 🐱

docs/api/cloudrootfs.md
```diff
@@ -13,44 +13,28 @@ cloud rootfs will package all the dependencies refers to the kubernetes cluster
...
-│   ├── init-kube.sh
-│   ├── init-registry.sh
-│   ├── init.sh
-│   └── kubelet-pre-start.sh
-└── statics # yaml files, sealer will render values in those files
-    └── audit-policy.yml
+├── plugins # plugins can run on some hooks, such as pre-init-host, post-install, see more in the plugins documentation
+│   └── disk_init_shell_plugin.yaml
+├── scripts # scripts can use all ClusterFile's env as Linux env variables
+│   ├── init-kube.sh # initialize kube* binaries on target hosts
```
Starnop, line 30: Could you please add more information to the scripts, including input, output, and functions should be included?

docs/api/cloudrootfs.md
```diff
@@ -13,44 +13,28 @@ cloud rootfs will package all the dependencies refers to the kubernetes cluster
 │   ├── kubectl
 │   ├── kubelet
 │   ├── nerdctl
+│   ├── kubelet-pre-start.sh
+│   ├── helm
```
Starnop, line 17: What if a dual-architecture cluster image, such as support for both amd64 and arm64?

I think we should have a relatively complete example.

docs/api/cloudrootfs.md
Starnop: Agree with that. It's weird to have a bunch of shell scripts in a directory called bin

PR #1924, opened 2022-12-20
chore: rename variables

kakaZhou719:
1. rename some variables
2. use `net.JoinHostPort` to join host port


Starnop (approved):
LGTM

PR #1925, opened 2022-12-20
support remove taint

Stevent-fei:


VinceCui:
We should not do delete operation when using clusterfile, if user want delete taints, he can use kubectl directly.

PR #1928, opened 2022-12-21
chore: update concurrent for some logic

Starnop:

We need to be more explicit about when we should use concurrency, and when using concurrency actually increases our chances of making mistakes. In this PR, I update these things:
1. set the concurrency of pull image is set to 1 because the bottleneck of downloading images is often in the network rather than computing resources
2. copy the registry to other master nodes, we will copy lots of small files and the bottleneck of copy registry is not in the network so we can copy multi nodes concurrently


kakaZhou719 (approved):
LGTM

PR #1931, opened 2022-12-22
feature: check image blob save and retry if not exist

Starnop:

check if the blob files for image is complete and retry to get if not


kakaZhou719 (approved):
LGTM

PR #1937, opened 2022-12-27
modify lvscare name

Stevent-fei:

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -40,6 +40,7 @@ const (
 	KubeLvsCareStaticPodName = "kube-lvscare"
 	RegLvsCareStaticPodName  = "reg-lvscare"
 	StaticPodDir             = "/etc/kubernetes/manifests"
+	LvsCareRepoAndTag        = "sealerio/lvscare:v1.1.3-beta.8"
 )
 
 // API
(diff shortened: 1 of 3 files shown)


kakaZhou719:
how about make it as a const value , replace it  including registry lvscare and k8s runtime lvscare

common/common.go
```diff
@@ -40,6 +40,7 @@ const (
 	KubeLvsCareStaticPodName = "kube-lvscare"
 	RegLvsCareStaticPodName  = "reg-lvscare"
 	StaticPodDir             = "/etc/kubernetes/manifests"
+	LvsCare                  = "sealerio/lvscare:v1.1.3-beta.8"
```
kakaZhou719, line 43: LvsCareRepoAndTag

pkg/registry/local.go
```diff
@@ -162,7 +162,7 @@ func (c *localConfigurator) configureLvs(registryHosts, clientHosts []net.IP) er
 	}
 
 	//todo should make lvs image name as const value in sealer repo.
-	lvsImageURL := fmt.Sprintf("%s/sealer/lvscare:v1.1.3-beta.8", net.JoinHostPort(c.Domain, strconv.Itoa(c.Port)))
+	lvsImageURL := filepath.Join(net.JoinHostPort(c.Domain, strconv.Itoa(c.Port)), common.LvsCare)
```
kakaZhou719, line 165: in cluster runtime, we also need to replace this value.

Starnop (approved):
LGTM

PR #1945, opened 2023-01-04
feat: support ipvs vip overwriting 

kakaZhou719:

we can overwrite the default ipvs VIP through  cluster env.


Starnop (approved):
LGTM

PR #1954, opened 2023-01-05
[WIP]Add image list description

Stevent-fei:
image

### DIFF
--- a/cmd/sealer/cmd/alpha/search.go
+++ b/cmd/sealer/cmd/alpha/search.go
@@ -17,9 +17,11 @@ package alpha
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	reference2 "github.com/distribution/distribution/v3/reference"
 	"github.com/olekukonko/tablewriter"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


cmd/sealer/cmd/alpha/search.go
```diff
@@ -70,10 +71,15 @@ func NewSearchCmd() *cobra.Command {
 				}
 				tags, err := repo.Tags(context.Background()).All(context.Background())
 				if err != nil {
-					return err
+					return fmt.Errorf("repository %s does not exist, err:%v", imgName, err)
```
Starnop, line 74: errors.Wrapf(err,"repository %s does not exist", imgName)
 
 maybe is better here

cmd/sealer/cmd/alpha/search.go
```diff
@@ -70,10 +71,15 @@ func NewSearchCmd() *cobra.Command {
...
 				}
 				for _, tag := range tags {
-					table.Append([]string{named.String(), tag})
+					if strings.Contains(tag, "-") {
```
Starnop, line 77: I don't think it's a good idea to get network plugins by image tag

PR #1955, opened 2023-01-05
support merge flags when use clusterfile for run and apply

kakaZhou719:

1. support merge flags when use clusterfile for run and apply
2. bugfix: no need to specify master ip when run app image
3. optimize: each cmd use their own flags


VinceCui (commented):
LGTM

VinceCui (approved):

PR #1959, opened 2023-01-06
support specify appname for config

kakaZhou719:

`ConfigSpec.path` is where the `ConfigSpec.Data` write on, it could be the relative path of rootfs (manifests/mysql.yaml). Or work with APPName to form a complete relative path(application/apps/{APPName}/mysql.yaml)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -22,14 +22,13 @@ import (
 	"strings"
 
 	"github.com/imdario/mergo"
-
+	"github.com/sealerio/sealer/pkg/rootfs"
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -21,6 +21,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/sealerio/sealer/pkg/rootfs"
```
Starnop, line 24: Reorganize the go import group

Starnop (approved):
LGTM

PR #1960, opened 2023-01-06
chore: add container image list for image spec

Starnop:

add container image list for sealer image spec

the end users can get container image list by `sealer inspect` and the developer can get the container image list by  the key `sealer.image.container.images` from image annotations

### Special notes for reviews
after that, we also need to do the following:
1. parse and store the container image list when the build phase
2. add a flag for `sealer inspect` to decide whether to output the list of images because of container image list could be a huge number

### DIFF
--- a/pkg/define/image/v1/sealer_image_extension.go
+++ b/pkg/define/image/v1/sealer_image_extension.go
@@ -24,7 +24,9 @@ import (
 )
 
 const (
-	SealerImageExtension = "sealer.image.extension"
+	SealerImageExtension          = "sealer.image.extension"
+	SealerImageContainerImageList = "sealer.image.container.images"
+
 	// Kube image type
 	KubeInstaller = "kube-installer"
 	AppInstaller  = "app-installer"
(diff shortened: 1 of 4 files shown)


pkg/define/image/v1/sealer_image_extension.go
```diff
@@ -49,6 +51,23 @@ type ImageSpec struct {
 	OCIv1 ociv1.Image `json:"ociv1"`
 
 	ImageExtension
+
+	// ContainerImageList the container image list contained in the sealer image
+	//
+	// TODO: populate this value during the build phase
+	ContainerImageList []*ContainerImage `json:"containerImageList,omitempty"`
```
kakaZhou719, line 58: how about:   map[AppName][]string{images}

Starnop: I don't think so. Here's a case, a container image may not belong to any app.

kakaZhou719 (approved):
LGTM

PR #1975, opened 2023-01-14
Feat/support containerd

VinceCui:


cmd/sealer/cmd/cluster/run.go
```diff
@@ -253,9 +253,10 @@ func createNewCluster(imageEngine imageengine.Interface, cf clusterfile.Interfac
 	}
 
 	runtimeConfig := &clusterruntime.RuntimeConfig{
-		Distributor: distributor,
-		ImageEngine: imageEngine,
-		Plugins:     plugins,
+		Distributor:            distributor,
+		ImageEngine:            imageEngine,
+		Plugins:                plugins,
+		ContainerRuntimeConfig: cluster.Spec.ContainerRuntime,
```
kakaZhou719, line 259: could we get ContainerRuntime from image spec(image extension) as default value, using cluster.Spec.ContainerRuntime  to ovewrite it ?

pkg/runtime/kubernetes/join_masters.go
```diff
@@ -102,6 +102,13 @@ func (k *Runtime) joinMasters(newMasters []net.IP, master0 net.IP, kubeadmConfig
 			return err
 		}
 
+		if err = k.infra.CmdAsync(m, "mv /etc/kubernetes/manifests/kube-scheduler.yaml /tmp/ && mv /tmp/kube-scheduler.yaml /etc/kubernetes/manifests/"); err != nil {
```
kakaZhou719, line 105: Can you pls add a description to this code? its hard to understand  -:)

pkg/container-runtime/installer.go
```diff
@@ -44,39 +49,52 @@ type Installer interface {
...
 	CertsDir       string
 	ConfigFilePath string
 }
 
-func NewInstaller(conf Config, driver infradriver.InfraDriver) (Installer, error) {
-	if conf.Type == "docker" {
-		return &DockerInstaller{
+func NewInstaller(conf v2.ContainerRuntimeConfig, driver infradriver.InfraDriver) (Installer, error) {
+	switch conf.Type {
+	case "docker", "":
```
kakaZhou719, line 62: could we make "docker","containerd" as a const ? or make it as a valid map["container runtime type"]struct{}, so if another type not found, we cpuld give more detailed error message.

pkg/container-runtime/installer.go
```diff
@@ -44,39 +49,52 @@ type Installer interface {
...
+		}
+		ret.Info.CgroupDriver = DefaultCgroupDriver
+		if cd, ok := ret.envs[CgroupDriverArg]; ok && cd != nil {
+			ret.Info.CgroupDriver = cd.(string)
+		}
 
-	return nil, fmt.Errorf("please enter the correct container type")
+		return ret, nil
+	default:
+		return nil, fmt.Errorf("please enter the correct container type")
```
kakaZhou719, line 98: return nil, fmt.Errorf("invalid container runtime type" )

cmd/sealer/cmd/cluster/run.go
VinceCui: i think after we add ContainerRuntime info in  image spec, then modify here.

pkg/container-runtime/default.go
```diff
@@ -19,37 +19,47 @@ import (
...
 	return nil
 }
 
-func (d DockerInstaller) GetInfo() (Info, error) {
+func (d *DefaultInstaller) GetInfo() (Info, error) {
 	return d.Info, nil
 }
+
+func (d *DefaultInstaller) getInstallScriptName() string {
+	return fmt.Sprintf("%s.sh", d.Type)
```
kakaZhou719, line 60: if container runtime shell file renamed ,is the corresponding rootfs files also need to renamed ?, how to ensure compatibility for the old sealer image.

VinceCui: Let's make a statement

VinceCui: the cluster image provider should make sure the script for installing container runtime use the same name with container runtime type.

pkg/clusterfile/decoder.go
```diff
@@ -165,29 +165,63 @@ func checkAndFillCluster(cluster *v2.Cluster) error {
...
+	}
+	cluster.Spec.Env = append(cluster.Spec.Env, fmt.Sprintf("%s=%s", common.EnvRegistryPort, portStr))
 	registryURL := net.JoinHostPort(regConfig.Domain, strconv.Itoa(regConfig.Port))
 	if regConfig.Port == 0 {
 		registryURL = regConfig.Domain
 	}
 	cluster.Spec.Env = append(cluster.Spec.Env, fmt.Sprintf("%s=%s", common.EnvRegistryURL, registryURL))
 
+	if cluster.Spec.ContainerRuntime.Type != "" {
+		cluster.Spec.Env = append(cluster.Spec.Env, fmt.Sprintf("%s=%s", common.EnvContainerRuntime, cluster.Spec.ContainerRuntime.Type))
```
starComingup, line 223: @VinceCui I think read container runtime type by clusterfile is a little complex. if user didn't specify CRI in clusterfile, What will happended? Then if user specify a wrong CRI type in clusterfile which different from clusterimage had, what will happeded?
Here is my suggestion: 
now we have `LABEL` in Kubefile, so we can get CRI info from it. When user use offical image, the CRI info will natural included. If CRI info is not specified in Kubefile, we use docker as a default CRI and give user a hint, if CRI info is wrong, we hint the default CRI info and return error for user. 
RFC @Starnop

VinceCui: Thanks for the review. 

The first question, if user didn't specify CRI, the default value now is docker, but it may be changed to containerd in future.

The second question, it is no problem to specify the default container runtime by LABEL in the build phase. This does not conflict with the ability to specify the container runtime at runtime. But consider a scenario, what if the image supports multiple container runtimes? ? At this time, it is necessary to allow users to specify at runtime.

感谢review。
第一个问题，如果置为空，默认为docker，但后面可能会改成默认containerd。
第二个问题，在build阶段里LABEL指定默认的容器运行时是没有问题的，这和运行时可以指定容器运行时也是不冲突的，但是考虑一个场景，如果该镜像支持多个容器运行时呢？此时允许使用者在运行时可以指定就是必要的了。

kakaZhou719 (approved):
LGTM

PR #1985, opened 2023-01-17
fix sealer alpha mount bug

Stevent-fei:
If the image is too small, the container directory will be mounted without being created, so you need to wait for 1s


Starnop (approved):
LGTM

PR #2002, opened 2023-02-02
add getImageLabels func, to get CRI and cluster runtime info.

starComingup:

We need to use image labels when we install different CRI、CSI、CNI and so on, so we need to get image labels info to help us understand which were specified by clusterimage.

NONE

### Special notes for reviews
If I did this correctly. I will use this function in sealer run.go.

### DIFF
--- a/pkg/imageengine/buildah/get_image_info.go
+++ b/pkg/imageengine/buildah/get_image_info.go
@@ -39,3 +39,21 @@ func (engine *Engine) GetImageAnnotation(opts *options.GetImageAnnoOptions) (map
 	out := buildah.GetBuildInfo(builder)
 	return out.ImageAnnotations, nil
 }
+
+func (engine *Engine) GetImageLabels(opts *options.GetImageLabelsOptions) (map[string]string, error) {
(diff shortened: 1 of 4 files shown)


pkg/imageengine/buildah/get_image_info.go
```diff
@@ -39,3 +39,21 @@ func (engine *Engine) GetImageAnnotation(opts *options.GetImageAnnoOptions) (map
...
+	store := engine.ImageStore()
+	name := opts.ImageNameOrID
+
+	builder, err := openImage(ctx, engine.SystemContext(), store, name)
+	if err != nil {
+		return nil, err
+	}
+
+	out := buildah.GetBuildInfo(builder)
+	return out.OCIv1.Config.Labels, nil
```
Starnop, line 58: In fact, the labels that are defined in the `Kubefile` are stored in the ImageExtension

starComingup: Oops! so we just need to get ImageExtension to get labels, I will close this pr.

Issue #2011, opened 2023-02-03
[Feature] Support use CustomImageName to change TestImage for e2e test


zhy76:
## Issue Description

Type: *feature request*

### Describe what feature you want
I found that the previous image:docker.io/sealer-io/kubernetes:v1.22.15 has problems in the e2e test,
![image](https://user-images.githubusercontent.com/56665618/216630431-839e614f-603d-477a-8b11-e9331a0991f1.png)
![image](https://user-images.githubusercontent.com/56665618/216630491-036a1d98-c142-41a9-82f3-db038a0f5f69.png)
 but this image:docker.io/sealerio/kubernetes:v1-22-15-sealerio-2 is fine, it seems that the previous image is outdated. Change the test image requires changing all the relevant image names in the code, which is not very convenient. So, we can set an environment variable that changes only that variable to modify test image.

PR #2012, opened 2023-02-03
set CustomImageName to change TestImage and fix code

zhy76:

Change the test image requires changing all the relevant image names in the code, which is not very convenient. So, we can set an environment variable that changes only that variable to modify test image.
And after this PR, we can export an environment variable named IMAGE_NAME to modify all test image

#2011


kakaZhou719 (approved):
lgtm

Issue #2013, opened 2023-02-03
[Feature] Use client-go to interact with the cluster in container in e2e test


zhy76:
## Issue Description

Type: *feature request*

### Describe what feature you want
The previous e2e test uses ssh to obtain cluster node and pod information using the kubectl command in the container. Using client-go to get pods would better than using kubectl, kubectl would be for sure in official clusterimage,  when kubectl's version did not match, this would have an error in test stage.

In addition, if e2e test fails, pod logs cannot be obtained from the container, which makes debug very inconvenient, Especially in CI using github action, because it is like a black box. Therefore, client-go can be used to interact with the cluster in the container, and error logs can be printed for troubleshooting.
I think this is very important and helpful for e2e test.

Starnop:
Can't help but like 👍

PR #2014, opened 2023-02-03
use client-go to interact with the cluster in container and optimize code

zhy76:

After this PR, if e2e test failed, we can view the failed pod log, which is very helpful for us to debug.
for example:
image

#2013

### DIFF
--- a/.github/workflows/e2e-test-run.yml
+++ b/.github/workflows/e2e-test-run.yml
@@ -71,4 +71,6 @@ jobs:
           RegionID: ${{ secrets.RegionID }}
         if: ${{ github.event.comment.body == '/test run' || github.event.comment.body == '/test all' || github.event_name == 'push' }}
         run: |
+          # fix bug in kernal 5.12.2+:open /proc/sys/net/netfilter/nf_conntrack_max: permission denied, see: https://github.com/kubernetes-sigs/kind/issues/2240
+          sudo sysctl net/netfilter/nf_conntrack_max=131072
           ginkgo -v --focus="sealer run" test
--- a/test/testhelper/client/k8s/client.go
+++ b/test/testhelper/client/k8s/client.go
@@ -1,0 +1,285 @@
+// Copyright © 2023 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8s
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"path/filepath"
+	"regexp"
+
+	"github.com/sealerio/sealer/test/testhelper"
+
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	ReadyStatus = "Ready"
+	TRUE        = "True"
+	FALSE       = "False"
+)
+
+type Client struct {
+	client *kubernetes.Clientset
+}
+
+type NamespacePod struct {
+	Namespace v1.Namespace
+	PodList   *v1.PodList
+}
+
+type EventPod struct {
+	Reason    string
+	Message   string
+	Count     int32
+	Type      string
+	Action    string
+	Namespace string
+}
+
+func NewK8sClient(sshClient *testhelper.SSHClient) (*Client, error) {
+	kubeconfigPath := filepath.Join("/root", ".kube", "config")
+
+	data := testhelper.GetRemoteFileData(sshClient, kubeconfigPath)
+	reg := regexp.MustCompile(`server: https://(.*):6443`)
+	data = reg.ReplaceAll(data, []byte(fmt.Sprintf("server: https://%s:6443", sshClient.RemoteHostIP.String())))
+	config, err := clientcmd.RESTConfigFromKubeConfig(data)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to build kube config")
+	}
+
+	clientSet, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{
+		client: clientSet,
+	}, nil
+}
+
+func (c *Client) ListNodes() (*v1.NodeList, error) {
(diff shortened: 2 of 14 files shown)


.github/workflows/e2e-test-run.yml
```diff
@@ -71,4 +71,6 @@ jobs:
           RegionID: ${{ secrets.RegionID }}
         if: ${{ github.event.comment.body == '/test run' || github.event.comment.body == '/test all' || github.event_name == 'push' }}
         run: |
+          # fix:open /proc/sys/net/netfilter/nf_conntrack_max: permission denied, see: https://github.com/kubernetes-sigs/kind/issues/2240
+          sudo sysctl net/netfilter/nf_conntrack_max=131072
```
Starnop, line 75: Also note the operating system and kernel that this configuration is for

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,276 @@
...
+	Type      string
+	Action    string
+	Namespace string
+}
+
+func NewK8sClient(sshClient *testhelper.SSHClient) (*Client, error) {
+	kubeconfigPath := filepath.Join("/root", ".kube", "config")
+
+	data := testhelper.GetFileDataRemote(sshClient, kubeconfigPath)
+	data = bytes.ReplaceAll(data, []byte("apiserver.cluster.local"), []byte(sshClient.RemoteHostIP.String()))
```
Starnop, line 62: Please do not rely on the `cluster.local` domain which is configurable. 

Maybe it's better to  replace `server: https://(.*):6443` to `server: https://sshClient.RemoteHostIP.String():6443`

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,276 @@
...
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"path/filepath"
+
+	"github.com/pkg/errors"
```
Starnop, line 25: go import group

the imported modules should be grouped look like this:
1. standard library modules
2. current project modules
3. 3rd party modules

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,276 @@
...
+
+func (c *Client) GetNotReadyPodEvent() (map[string][]EventPod, error) {
+	namespacePodList, err := c.ListAllNamespacesPods()
+	if err != nil {
+		return nil, err
+	}
+	result := make(map[string][]EventPod)
+	for _, podNamespace := range namespacePodList {
+		for _, pod := range podNamespace.PodList.Items {
+			if err := c.getPodReadyStatus(pod); err != nil {
```
Starnop, line 206: if err := c.getPodReadyStatus(pod); err == nil {
  continue
}

In programming, we need to pay attention to two things:
1. return fast
2. avoid too many depth

zhy76:
gosec / golang-security-action shows：
image
It has nothing to do with the code I changed, seems like a bug? How do I solve these problem?

kakaZhou719:
> gosec / golang-security-action shows： image It has nothing to do with the code I changed, seems like a bug? How do I solve these problem?

its gosec lint, i will fix it today.

zhy76:
PTAL

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,284 @@
...
+	events, err := c.client.CoreV1().
+		Events(nameSpace).
+		List(context.TODO(), metav1.ListOptions{FieldSelector: "involvedObject.name=" + podName, TypeMeta: metav1.TypeMeta{Kind: "Pod"}})
+	if err != nil {
+		return nil, err
+	}
+	return events.Items, nil
+}
+
+func (c *Client) getPodReadyStatus(pod v1.Pod) error {
```
justadogistaken, line 194: I think this function only checks if the pod is ready, so returning true/false would be good. What do you think?

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,284 @@
...
+			continue
+		}
+		if !pod.Status.ContainerStatuses[0].Ready {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (c *Client) GetPodLog(nameSpace, podName string) (string, error) {
```
justadogistaken, line 164: "nameSpace" => "namespace"

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,284 @@
...
+		_ = podLogs.Close()
+	}()
+
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, podLogs)
+	if err != nil {
+		return "", err
+	}
+	str := buf.String()
+	return str, nil
```
justadogistaken, line 181: return buf.String(), nil

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,284 @@
...
+			Namespace: ns,
+			PodList:   pods,
+		}
+		namespacePodList = append(namespacePodList, &namespacePod)
+	}
+
+	return namespacePodList, nil
+}
+
+func (c *Client) ListKubeSystemPodsStatus() (bool, error) {
```
justadogistaken, line 143: Actually, I don't think the function name fits what the function truly does.
The function name makes the callers feel it will return a list(KubeSystemPodsStatus), but it checks if all pods of "KubeSystem" are ready.

test/testhelper/client/k8s/client.go
zhy76: Make sense

test/testhelper/utils.go
```diff
@@ -162,6 +162,14 @@ func GetFileDataLocally(filePath string) string {
 	return result
 }
 
+// GetFileDataRemote get file data from remote
+func GetFileDataRemote(sshClient *SSHClient, filePath string) []byte {
```
Starnop, line 166: s/GetFileDataRemote/GetRemoteFileData

the result is FileData, not Remote, right?

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,283 @@
...
+	Action    string
+	Namespace string
+}
+
+func NewK8sClient(sshClient *testhelper.SSHClient) (*Client, error) {
+	kubeconfigPath := filepath.Join("/root", ".kube", "config")
+
+	data := testhelper.GetFileDataRemote(sshClient, kubeconfigPath)
+	reg := regexp.MustCompile(`server: https://(.*):6443`)
+	data = reg.ReplaceAll(data, []byte("server: https://"+sshClient.RemoteHostIP.String()+":6443"))
```
Starnop, line 64: use fmt.Sprinf to concatenate strings instead of +

test/testhelper/client/k8s/client.go
```diff
@@ -0,0 +1,283 @@
...
+
+// Check if all nodes are ready
+func (c *Client) CheckAllNodeReady() (bool, error) {
+	nodes, err := c.ListNodes()
+	if err != nil {
+		return false, err
+	}
+	for _, node := range nodes.Items {
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == ReadyStatus {
```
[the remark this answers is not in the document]
zhy76: optimized already

Starnop (approved):
LGTM

PR #2023, opened 2023-02-10
[bugfix]: add pre-check that If the cluster already exists

### LINKED ISSUE #2022: [sealer run]check if there is a running cluster, if exist, should not mount new image.
### What happen?

对于一个已经部署了集群镜像A的集群1，如果再次运行集群镜像B，那么sealer目前不会检查，然后直接mount了集群镜像B，可能就将镜像A的文件冲掉了。

For a cluster 1 that has already deployed cluster-image A, if cluster-image B is run again, the sealer will  directly mounts cluster image B without checking, and then may overwrite the file of image A.

### Relevant log output?

_No response_

我希望sealer run时，会提前检查是否有已经部署的集群镜像，如果有，且版本不一致，那么报错，如果版本一致，那么跳过。如果没有，正常安装

I hope that when the sealer runs, it will check whether there is a deployed cluster image. If there is, and the version is not equal, then throw an error. If the version is equal, then skip. If not, install normally

_No response_

_No response_

_No response_

_No response_

_No response_

- Install tools:
- Others:

Stevent-fei:

When the sealer runs, it will check in advance whether there is a deployed cluster image. If there is, and the version is inconsistent, then an error will be reported. If the version is consistent, then skip. If not, install normally

exist：
image

not exist：
image

Fixes #2022

### DIFF
--- a/cmd/sealer/cmd/cluster/run.go
+++ b/cmd/sealer/cmd/cluster/run.go
@@ -236,6 +236,11 @@ func runWithClusterfile(clusterFile string, runFlags *types.RunFlags) error {
 
 func runClusterImage(imageEngine imageengine.Interface, cf clusterfile.Interface, imageSpec *imagev1.ImageSpec, mode string) error {
 	cluster := cf.GetCluster()
+	//check cluster
+	checkClusterFile, err := utils.CheckClusterIsExists(cluster)
+	if checkClusterFile != "" {
+		return err
+	}
 	infraDriver, err := infradriver.NewInfraDriver(&cluster)
 	if err != nil {
 		return err
--- a/cmd/sealer/cmd/utils/cluster.go
+++ b/cmd/sealer/cmd/utils/cluster.go
@@ -26,12 +26,15 @@ import (
 	"github.com/sealerio/sealer/cmd/sealer/cmd/types"
 	"github.com/sealerio/sealer/common"
 	"github.com/sealerio/sealer/pkg/client/k8s"
+	"github.com/sealerio/sealer/pkg/clusterfile"
 	"github.com/sealerio/sealer/types/api/constants"
 	v1 "github.com/sealerio/sealer/types/api/v1"
 	v2 "github.com/sealerio/sealer/types/api/v2"
 	netutils "github.com/sealerio/sealer/utils/net"
 	strUtils "github.com/sealerio/sealer/utils/strings"
 )
+
+const clusterExists = "exist"
 
 func MergeClusterWithFlags(cluster v2.Cluster, mergeFlags *types.MergeFlags) (*v2.Cluster, error) {
 	if len(mergeFlags.CustomEnv) > 0 {
@@ -255,4 +258,16 @@ func GetClusterClient() *k8s.Client {
 		logrus.Warnf("try to new k8s client via default kubeconfig, maybe this is a new cluster that needs to be created: %v", err)
 	}
 	return nil
+}
+
+func CheckClusterIsExists(cluster v2.Cluster) (string, error) {
+	clusterFile, _, err := clusterfile.GetActualClusterFile()
+	if err == nil {
+		newClusterfile := clusterFile.GetCluster()
+		if cluster.Spec.Image != newClusterfile.Spec.Image {
+			return clusterExists, fmt.Errorf("the cluster image already exists, please uninstall the current cluster and install another version of the cluster image")
+		}
+		return clusterExists, fmt.Errorf("this cluster image has already been installed, please do not repeat the operation")
+	}
+	return "", nil
 }


cmd/sealer/cmd/cluster/run.go
```diff
@@ -115,6 +118,24 @@ func NewRunCmd() *cobra.Command {
...
+				file, err := os.Open(common.GetDefaultClusterfile())
+				if err != nil {
+					return err
+				}
+				context, err := ioutil.ReadAll(file)
+				if err != nil {
+					return err
+				}
+
+				if ok := strings.Contains(string(context), args[0]); !ok {
```
VinceCui, line 131: No need check if cluster image same, if not same, the user also can't run cluster.

VinceCui: Uer API to marshal it.

kakaZhou719:
@Stevent-fei i have two questions:

1. how to determine there is an existed cluster?
2. should we add this pre-check logic to all create process?

cmd/sealer/cmd/cluster/run.go
```diff
@@ -115,6 +117,30 @@ func NewRunCmd() *cobra.Command {
 					app, extension, nil, imageEngine, runFlags.Mode)
 			}
 
+			if exist := osi.IsFileExist(common.GetDefaultClusterfile()); exist {
```
kakaZhou719, line 120: not use file existing  to do this pre-check, try get  it from `clusterfile.NewClusterFile(nil)`,
 if not found, no cluster.
 if have ,determine whether the image name is the same one.

Starnop (approved):
LGTM

cmd/sealer/cmd/cluster/run.go
```diff
@@ -427,6 +433,19 @@ func installApplication(appImageName string, envs []string, app *v2.Application,
 	return nil
 }
 
+func checkClusterIsExists(cf clusterfile.Interface) error {
+	rawCluster := cf.GetCluster()
```
Starnop, line 437: rawCluster/desiredCluster/g

cmd/sealer/cmd/cluster/run.go
```diff
@@ -427,6 +433,19 @@ func installApplication(appImageName string, envs []string, app *v2.Application,
...
+	file, err := clusterfile.NewClusterFile(nil)
+	if err == nil {
```
Starnop, line 439: if err!=nil && match some regular，return nil

cmd/sealer/cmd/cluster/run.go
```diff
@@ -510,3 +517,16 @@ func prepareMaterials(infraDriver infradriver.InfraDriver, imageEngine imageengi
 	}
 	return nil
 }
+
+func checkClusterIsExists(cf clusterfile.Interface) error {
```
kakaZhou719, line 521: not good use err to identify the check result. i assume  we could return passed result and error.

kakaZhou719:
@Stevent-fei . it ok for me to do pre-check, such as cluster image . however , that is just one item need to check . it is better that if we add a check list to do this.

cmd/sealer/cmd/cluster/run.go
```diff
@@ -510,3 +518,16 @@ func prepareMaterials(infraDriver infradriver.InfraDriver, imageEngine imageengi
 	}
 	return nil
 }
+
+func checkClusterIsExists(cf clusterfile.Interface) (string, error) {
```
kakaZhou719, line 522: func checkClusterIsExists(cluster v2.cluster) (string, error) {

kakaZhou719, line 522: move this func to different file Or utils, and add UT test.

cmd/sealer/cmd/cluster/run.go
```diff
@@ -510,3 +518,16 @@ func prepareMaterials(infraDriver infradriver.InfraDriver, imageEngine imageengi
...
+	rawCluster := cf.GetCluster()
+	clusterFile, _, err := clusterfile.GetActualClusterFile()
+	if err == nil {
+		cluster := clusterFile.GetCluster()
+		if cluster.Spec.Image != rawCluster.Spec.Image {
+			return "exist", fmt.Errorf("the cluster image already exists, please uninstall the current cluster and install another version of the cluster image")
```
kakaZhou719, line 528: "exist" is not good, you should use const such as "SameImageName" Or something to identity it

cmd/sealer/cmd/cluster/run.go
```diff
@@ -235,6 +237,12 @@ func runWithClusterfile(clusterFile string, runFlags *types.RunFlags) error {
 }
 
 func runClusterImage(imageEngine imageengine.Interface, cf clusterfile.Interface, imageSpec *imagev1.ImageSpec, mode string) error {
+	//check cluster
+	checkClusterFile, err := checkClusterIsExists(cf)
+	if checkClusterFile == "exist" && err != nil {
```
kakaZhou719, line 242: why &&? Return err if it is not nil.

cmd/sealer/cmd/utils/cluster.go
```diff
@@ -256,3 +259,15 @@ func GetClusterClient() *k8s.Client {
 	}
 	return nil
 }
+
+func CheckClusterIsExists(cluster v2.Cluster) (string, error) {
+	clusterFile, _, err := clusterfile.GetActualClusterFile()
```
kakaZhou719, line 264: should retrun err fast

cmd/sealer/cmd/utils/cluster.go
```diff
@@ -256,3 +259,15 @@ func GetClusterClient() *k8s.Client {
...
+	if err == nil {
+		newClusterfile := clusterFile.GetCluster()
+		if cluster.Spec.Image != newClusterfile.Spec.Image {
+			return clusterExists, fmt.Errorf("the cluster image already exists, please uninstall the current cluster and install another version of the cluster image")
```
kakaZhou719, line 268: should not retrun err

Issue #1999, opened 2023-02-01
Support linux user with root privileges to install ClusterImage


VinceCui:
## Issue Description

Type: *feature request*

AFAK, Installing k8s does not necessarily require a root user, but a user with root privileges. There are some cases that the user don't want provide root user, so I think sealer need support this feature.

kakaZhou719:
sealer 0.9.2 is implemented this feature .

PR #2048, opened 2023-02-17
Feat/support nvidia docker and nonroot

VinceCui:
This PR support these features:
1. Support non root user use sudo to install.
2. If node to scale is alreay in current k8s cluster, just skip it.
3. Split NewClusterFile to NewClusterFile and GetActualClusterFile.
4. Support get customize cri socket path for each node.
5. Add upgrade-host plugin.

### DIFF
--- a/cmd/sealer/cmd/cluster/scale-up.go
+++ b/cmd/sealer/cmd/cluster/scale-up.go
@@ -66,13 +66,25 @@ func NewScaleUpCmd() *cobra.Command {
 				return fmt.Errorf("failed to parse ip string to net IP list: %v", err)
 			}
 
-			cf, err = clusterfile.NewClusterFile(nil)
+			cf, _, err = clusterfile.GetActualClusterFile()
 			if err != nil {
 				return err
 			}
 
 			cluster := cf.GetCluster()
-			if err = utils.ConstructClusterForScaleUp(&cluster, scaleUpFlags, scaleUpMasterIPList, scaleUpNodeIPList); err != nil {
+			client := utils.GetClusterClient()
+			if client == nil {
+				return fmt.Errorf("failed to get cluster client")
+			}
+
+			currentCluster, err := utils.GetCurrentCluster(client)
+			if err != nil {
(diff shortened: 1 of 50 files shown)


cubxxw:
#### resolve

Resolved issue: #1491、#1999

**Resolved remote read permission to `sshd_config` file:**
```bash
$ ls -al /etc/ssh/sshd_config
-rw------- 1 root root 3797 Feb 15 16:35 /etc/ssh/sshd_config
$ sudo chmod 644 /etc/ssh/sshd_config
```

<br>

#### existing problem

+ [ ] **Still needs to assign non-secret permissions to user sudo(Every node except master0):**
```bash
# echo "sealer ALL=(ALL) NOPASSWD: NOPASSWD: ALL" >> /etc/sudoers
```

+ [ ] **The rootless feature of [buildah](https://github.com/containers/buildah) is not addressed:**

cubxxw (approved):

cubxxw (approved):

cmd/sealer/cmd/cluster/scale-up.go
```diff
@@ -160,12 +161,6 @@ func scaleUpCluster(clusterImageName string, scaleUpMasterIPList, scaleUpNodeIPL
 		return err
 	}
 
-	//we need to save desired clusterfile to local disk temporarily.
-	//and will use it later to clean the cluster node if ScaleUp failed.
```
Starnop, line 164: it looks like we should not delete the code according to the comments

pkg/runtime/kubernetes/utils.go
```diff
@@ -53,6 +53,7 @@ const (
 )
 
 const (
+	GetCustomizeCRISocket   = "cat /etc/sealer-cri/socket-path"
```
Starnop, line 56: /etc/sealerio/cri/socket-path

cubxxw (approved):

kakaZhou719 (approved):
LGTM

PR #2070, opened 2023-02-28
feature: ensure the apply will always apply the applications

Starnop:

At present, if we re-apply a `Sealer Image` with the `kube-installer` type, it will return directly when no nodes are added.

For KxS runtime, maybe it looks good to add a pre-check and avoid some useless work because it's not allowed to update the cluster-related config. However, that's not right for the application,  the application config should be applied every time when `sealer apply` is executed.


build/kubefile/parser/utils.go
```diff
@@ -21,9 +21,10 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+
```
kakaZhou719, line 25: Format import ?

cmd/sealer/cmd/cluster/run.go
```diff
@@ -16,10 +16,13 @@ package cluster
 
 import (
 	"fmt"
-	"net"
 	"os"
 	"path/filepath"
 
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
```
kakaZhou719, line 24: ditto

cmd/sealer/cmd/cluster/apply.go
```diff
@@ -100,64 +102,31 @@ func NewApplyCmd() *cobra.Command {
...
-
-			infraDriver, err := infradriver.NewInfraDriver(&desiredCluster)
-			if err != nil {
-				return err
+				if err := applyClusterWithExisted(cf, client, imageEngine, imageSpec); err != nil {
+					return err
+				}
+				// NOTE: we should continue to apply application after the cluster is applied successfully
+				// And it's not needed to prepare the app file repeatedly
+				ignorePrepareAppMaterials = true
```
kakaZhou719, line 122: skipPrepareAppMaterials?

cmd/sealer/cmd/cluster/run.go
```diff
@@ -325,34 +327,9 @@ func loadPluginsFromImage(imageMountInfo []imagedistributor.ClusterImageMountInf
...
-	if err := distributor.DistributeRegistry(deployHosts, filepath.Join(infraDriver.GetClusterRootfsPath(), "registry")); err != nil {
-		return err
-	}
-
-	logrus.Infof("load image success")
-	return nil
-}
-
-func installApplication(appImageName string, envs []string, app *v2.Application, extension imagev1.ImageExtension, configs []v1.Config, imageEngine imageengine.Interface, mode string) error {
+func installApplication(appImageName string, envs []string, app *v2.Application,
```
kakaZhou719, line 330: too much func args, should we new an option including those args?

pkg/cluster-runtime/apps.go
```diff
@@ -55,32 +56,44 @@ func (i *AppInstaller) Install(master0 net.IP, cmds []string) error {
...
+	}
+	return nil
+}
+
+func (i AppInstaller) Install(master0 net.IP, launchCmds []string,
+	installMode string, ignorePrepareMaterials bool) error {
+	if installMode == common.ApplyModeLoadImage {
+		return LoadToRegistry(i.infraDriver, i.distributor)
+	}
+	if !ignorePrepareMaterials {
```
kakaZhou719, line 89: rename it to `prepareMaterials`, it could make code more understandable .

kakaZhou719 (approved):
LGTM

Issue #2001, opened 2023-02-01
[Task List] 0.9.2 Release - 20230228


Starnop:
## Issue Description

Type: *feature request*

### Universal capability
- [x] support multi-arch for sealer image
- [ ] support e2e test #1966
   - [x] support e2e test CI
   - [ ] test coverage reaches 30 percent
   - [x] support running e2e test locally
- [ ] Performance, disk usage optimization
   - [ ]  use image cache when build stage

### APP Image Related
- [x] support image policy with Kyverno
- [x]   continue to improve APP,APPCMDS instruction for Kubefile and related config for Clusterfile
- [ ] completely decouple the app image from the cluster image
  - [ ] run app image on any declared supported versions of the kubernetes cluster 
- [x] support for concurrent deployment of applications 
- [x] https://github.com/sealerio/sealer/issues/2017
### Cluster Image Related
- [x] support upgrade kubernetes cluster
- [x] support Linux user with root privileges to install ClusterImage  #1999 
- [x] support more container runtime
  - [x] support containerd
- [x] support more cluster runtime
  - [x] support k0s cluster runtime

### Bugfix
- [x] https://github.com/sealerio/sealer/issues/2000 @Stevent-fei 
- [x] https://github.com/sealerio/sealer/issues/1998 @Stevent-fei 
- [x] https://github.com/sealerio/sealer/issues/1995 @Starnop 
- [x] https://github.com/sealerio/sealer/issues/1994 @Stevent-fei
- [x] https://github.com/sealerio/sealer/issues/1993 @kakaZhou719 
- [x] https://github.com/sealerio/sealer/issues/1989 @kakaZhou719 
- [x] https://github.com/sealerio/sealer/issues/1986 @kakaZhou719

Issue #1940, opened 2022-12-30
[proposal] How to implement image storage for sealer using Skopeo


jsparter:
### background:
Now we always pull application images online at build stage, even if there are images in docker-daemon locally. And we pull blobs of images and manage all files by hands which is so inelegant. As #1874 said，we need to load offline images at build stage.

[Skopeo](https://github.com/containers/skopeo) is a image transporter, it can transport series of types of image from one to another, and `skopeo copy` ability can help us to implement load offline images. [Skopeo operates on the following image and repository types:](https://github.com/containers/skopeo#:~:text=Skopeo%20operates%20on,Specification%22%20at%20path.) `containers-storage:docker-reference`、`dir:path`、`docker://docker-reference`、`docker-archive:path[:docker-reference]`、`docker-daemon:docker-reference`、`oci:path:tag`. Skopeo can transport images between any two types above:
```
skopeo copy containers-storage:docker.io/library/alpine:3.13 docker://sea.hub:5000/library/alpine:3.13
skopeo copy docker-daemon:ubuntu:rolling docker://sea.hub:5000/library/ubuntu:rolling

```

`docker://docker-reference`、 `docker-daemon:docker-reference` and `oci:path:tag` these three image types may match our needs. But, each image type mentioned above has different structure,  `docker://docker-reference` needs there is a registry service running on. 

So, there are three solutions tu implement image store i figured:
1. Running a registry in build stage, and use `skopeo copy` ability to transport images to registry. It can help us fully make use of Skopeo, and completely offload loading image capability to Skopeo. Obviously, this approach relies on the Docker and registry containers, and brings complexity to the build stage.
2. Partly make use of `skopeo copy` to connect docker-daemon and read local images, then still manage all files by hands. It's hard to implement, and still inelegant. In the long run, if the Registry store format changes, our maintenance costs will be high.
3. Load images to a middle type at build stage using `skopeo copy` ability, and load middle type image to registry at run stage affter registry is running using `skopy copy` directly. For example, use `oci:path:tag` as middle type, so there are two step as follow：
    ```
    skopeo copy docker-daemon:docker.io/library/alpine:3.13 oci:$rootfs/ociimages:library/alpine:3.13
    skopeo copy oci:$rootfs/ociimages:library/alpine:3.13 docker://sea.hub:5000/library/alpine:3.13
    ```
[...]

PR #2071, opened 2023-03-01
feature: add oci registry

jsparter:
Already implement scenario1、2 of #1940. This pr needs to work with [basefs pr 40](https://github.com/sealerio/basefs/pull/40)
Now we can use image cache when build stage by specify registryType as "oci" in json file: $HOME/.sealer/config，and run with new basefs after [basefs pr 40](https://github.com/sealerio/basefs/pull/40)
config file example:
```
{
  "registryType": "oci"
}
```

Part of #1940 、#2001


pkg/define/config.go
```diff
@@ -0,0 +1,56 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/sealerio/sealer/common"
+	osi "github.com/sealerio/sealer/utils/os"
+)
+
+type SealerConfig struct {
```
kakaZhou719, line 28: i agree with you that we need a config file to configure the behavior of the sealer, such as build, run.  so, could we make a discuss together about the format of this file ? @Starnop.

pkg/image/save/skopeo/copy.go
```diff
@@ -0,0 +1,261 @@
...
+	// signIdentity used for sign
+	var signIdentity reference.Named = nil
+	//if opts.signIdentity != "" {
+	//	signIdentity, err = reference.ParseNamed(opts.signIdentity)
+	//	if err != nil {
+	//		return fmt.Errorf("Could not parse --sign-identity: %v", err)
+	//	}
+	//}
+
+	// 7. MUST，好像这里可以设置架构，需要测试
```
kakaZhou719, line 113: translate it?

pkg/image/save/oci_save.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (is *OCIImageSaver) save(imageType string, namds []Named, creds string, dir string) error {
+	for _, namd := range namds {
+		srcImageName := getImageNameWithStorageType(imageType, namd, "")
+		destImageName := getImageNameWithStorageType(skopeo.OciPath, namd, dir)
+		if err := getDestDir(destImageName); err != nil {
+			return err
+		}
+		err := skopeo.Copy(srcImageName, destImageName, creds)
```
fengxsong, line 162: This approach will make it impossible to reuse layers that may be duplicated in each image while building new oci image.

PR #2079, opened 2023-03-06
chore: optimize something

Starnop:

1. optimize the error log with the reason for failing to pull the `Seaelr Image`
2. we should expose any interface definitions including the constants we use


kakaZhou719 (approved):
LGTM

PR #2081, opened 2023-03-06
add upgrade cmd

wizardlyk:
change upgrade mode to cmd, support upgrade with cluster file

### DIFF
--- a/cmd/sealer/cmd/types/types.go
+++ b/cmd/sealer/cmd/types/types.go
@@ -87,3 +87,7 @@ type MergeFlags struct {
 	// override default APPNames of clusterimage
 	AppNames []string
 }
+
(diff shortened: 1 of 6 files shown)


cmd/sealer/cmd/types/types.go
```diff
@@ -87,3 +87,16 @@ type MergeFlags struct {
 	// override default APPNames of clusterimage
 	AppNames []string
 }
+
+type UpgradeFlags struct {
+	ClusterFile string
+	Masters     string
```
kakaZhou719, line 93: do we need those flags for upgrde?

wizardlyk: no, deleted

VinceCui (approved):
LGTM

PR #2082, opened 2023-03-06
fix clusterfile

Stevent-fei:

image
add ssh information when running the cluster


kakaZhou719:
@Stevent-fei ,if it’s is a bugfix,  please describe it as more as possible.

PR #2113, opened 2023-03-15
delete mount e2e test

Stevent-fei:
delete mount e2e test

unable to read, write and mount files due to insufficient container permissions.

I have recorded this problem, and I will try to solve it later.


kakaZhou719 (approved):
LGTM

PR #2116, opened 2023-03-15
bugfix: fix apply

Starnop:

Fix the following two issues:
1. apply with mode `loadImage` will also launch the app
2. apply with no node scale will launch the apps  and do not distribute the image rootfs

### DIFF
--- a/cmd/sealer/cmd/cluster/apply.go
+++ b/cmd/sealer/cmd/cluster/apply.go
@@ -117,7 +117,7 @@ func NewApplyCmd() *cobra.Command {
 
 			// NOTE: in some scenarios, we do not need to prepare the app file repeatedly,
 			// such as the cluster and the apps in the same image
-			var ignorePrepareAppMaterials bool
+			var skipPrepareAppMaterials bool
 			// ensure that the cluster reaches the final state firstly
 			if imageSpec.ImageExtension.Type == imagev1.KubeInstaller {
 				client := utils.GetClusterClient()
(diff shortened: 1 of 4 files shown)


cmd/sealer/cmd/cluster/apply.go
```diff
@@ -127,12 +127,13 @@ func NewApplyCmd() *cobra.Command {
 					return applyClusterWithNew(cf, applyMode, imageEngine, imageSpec)
 				}
 
-				if err := applyClusterWithExisted(cf, client, imageEngine, imageSpec); err != nil {
+				updated, err := applyClusterWithExisted(cf, client, imageEngine, imageSpec)
```
kakaZhou719, line 130: updated is too large to understand .  Or i thought should we move those run app logic  to applyClusterWithExisted.

cmd/sealer/cmd/cluster/apply.go
```diff
@@ -127,12 +127,13 @@ func NewApplyCmd() *cobra.Command {
 					return applyClusterWithNew(cf, applyMode, imageEngine, imageSpec)
```
kakaZhou719, line 127: applyClusterFileWithNew

cmd/sealer/cmd/cluster/apply.go
kakaZhou719, line 130: applyClusterFileWithExisted ?

cmd/sealer/cmd/cluster/apply.go
```diff
@@ -127,12 +127,13 @@ func NewApplyCmd() *cobra.Command {
...
 
-				if err := applyClusterWithExisted(cf, client, imageEngine, imageSpec); err != nil {
+				updated, err := applyClusterWithExisted(cf, client, imageEngine, imageSpec)
+				if err != nil {
 					return err
 				}
 				// NOTE: we should continue to apply application after the cluster is applied successfully
 				// And it's not needed to prepare the app file repeatedly
-				ignorePrepareAppMaterials = true
+				ignorePrepareAppMaterials = updated
```
kakaZhou719, line 136: SkipPrepareAppMaterials ?

cmd/sealer/cmd/cluster/apply.go
Starnop: in fact, it‘s just the part relevant to the cluster.

And as for what you said about the app, that's really unreasonable. And I have added a comment `TODO: decouple the cluster installation and application installation`, and we should fix it in a separate PR.

cmd/sealer/cmd/cluster/apply.go
Starnop: ditto

cmd/sealer/cmd/cluster/apply.go
Starnop: DONE

cmd/sealer/cmd/cluster/apply.go
Starnop: I have changed `updated` to `clusterUpdated`.

As with other comments, we should split run/apply to these steps:
1. distribute rootfs
2. apply/run cluster
3. app/run application

And then, something just like `clusterUpdated`, `skipPrepareAppMaterials ` will not be needed.

And we should do these code optimization ASAP.

kakaZhou719 (approved):
LGTM

PR #2121, opened 2023-03-17
move login to image module

kakaZhou719:
1. move login to image module
2. rename ginkgo image `Describe` same with `focus`


Stevent-fei (approved):
LGTM

PR #2122, opened 2023-03-17
fix get secret none

zhy76:
fix get secret none because pull_request event can't get github SECERT, this will cause sealer login failed.


kakaZhou719 (approved):
LGTM

PR #2135, opened 2023-03-22
Support set node name as node ip using env UseIPasNodeName.

VinceCui:


pkg/runtime/kubernetes/utils.go
```diff
@@ -136,6 +136,9 @@ func (k *Runtime) Command(name CommandType) (string, error) {
 	if !ok {
 		return "", fmt.Errorf("failed to get kubeadm command: %v", cmds)
 	}
+	if nodeNameOverride != "" {
+		v = fmt.Sprintf("%s --node-name %s", v, nodeNameOverride)
```
kakaZhou719, line 140: is this flag also valid to `initMaster ` ?

kakaZhou719 (approved):
LGTM

PR #2138, opened 2023-03-24
support set tmp dir for load image

kakaZhou719:

we could support set tmp dir for load image. In this way, it is possible to separate the temporary directory from the system disk

Example as below :

`sealer load -i myimage.tar --tmp-dir /my/tmp/disk`


Starnop (approved):
LGTM

PR #2140, opened 2023-03-24
feat: support env rendering for files in application/apps 

kakaZhou719:

1. add  `ApplicationConfig.Env` for supporting app env render.
2. move out all app files modifying func to file.go

Clusterfile:

```yaml
apiVersion: sealer.io/v2
kind: Cluster
metadata:
  name: my-cluster
spec:
  hosts:
  - ips:
    - 172.16.26.165
    roles:
    - master
  image: localhost/abc:v1
  ssh:
    passwd: Seadent123
    pk: /root/.ssh/id_rsa
    port: "22"
    user: root
status: {}
---
kind: Application
metadata:
  name: my-apps
spec:
  launchApps:
    - shell
    - yamlapp
  configs:
    - name: yamlapp
      env:
        - gk2=testvalue
        - imageName=new-image-name-for-render
        - k2=v2
    - name: shell
      files:
        - path: rediscert
          strategy: "overwrite"
          data: |
            redis-user: root
            redis-passwd: xxx
```

### DIFF
--- a/pkg/application/files.go
+++ b/pkg/application/files.go
@@ -1,0 +1,119 @@
+// Copyright © 2023 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package application
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/imdario/mergo"
+	"github.com/sealerio/sealer/pkg/env"
+	v2 "github.com/sealerio/sealer/types/api/v2"
+	osUtils "github.com/sealerio/sealer/utils/os"
+	"gopkg.in/yaml.v3"
+)
+
+func newFileProcessor(appFile v2.AppFile) (FileProcessor, error) {
+	switch appFile.Strategy {
+	case v2.OverWriteStrategy:
+		return overWriteProcessor{appFile}, nil
+	case v2.MergeStrategy:
+		return mergeProcessor{appFile}, nil
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/application/v2app.go
```diff
@@ -238,8 +238,15 @@ func NewV2Application(app *v2.Application, extension imagev1.ImageExtension) (In
 			v2App.appLaunchCmdsMap[name] = launchCmds
 		}
 
-		// initialize app files
+		// add app env
+		v2App.appEnvMap[name] = strUtils.ConvertStringSliceToMap(config.Env)
```
Starnop, line 242: it will panic here if the appEnvMap is nil

utils/maps/maps.go
```diff
@@ -33,3 +22,29 @@ func ConvertToSlice(m map[string]string) []string {
...
+	for srcEnvKey, srcEnvValue := range src {
+		if _, ok := dst[srcEnvKey]; ok {
+			continue
+		}
+		dst[srcEnvKey] = srcEnvValue
+	}
+	return dst
+}
+
+func Copy(origin map[string]string) map[string]string {
```
Starnop, line 40: also add unit test ?

pkg/application/v2app.go
```diff
@@ -167,6 +166,7 @@ func NewV2Application(app *v2.Application, extension imagev1.ImageExtension) (In
 		globalCmds:          extension.Launch.Cmds,
 		appLaunchCmdsMap:    map[string][]string{},
 		appRootMap:          map[string]string{},
+		appEnvMap:           map[string]map[string]string{},
```
Starnop, line 169: should we init the appEnvMap here to avoid the map objects with a value of nil being used? same with the others

pkg/application/files.go
```diff
@@ -0,0 +1,115 @@
...
+	return nil
+}
+
+//envRender :this will render the FilePath with the Values.
+type envRender struct {
+	envData map[string]string
+}
+
+func (e envRender) Process(appRoot string) error {
+	return env.RenderTemplate(appRoot, e.envData)
```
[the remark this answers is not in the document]
kakaZhou719: > what will occurs if the e.envData is nil

yes, should have a validation firstly , if  the `e.envData` is nil, just return

utils/maps/maps.go
kakaZhou719: ok ,i will add UT test .

Starnop (approved):
LGTM

PR #2175, opened 2023-04-10
bugfix: use os.Create to write env rendered file

kakaZhou719:
`os.OpenFile(strings.TrimSuffix(path, templateSuffix), os.O_CREATE|os.O_RDWR, os.ModePerm)`

if file is exits, without O_TRUNC mod, will leads to uncontrollable results. so we do better use `os.Create` to write rendered file.


Stevent-fei (approved):
LGTM

PR #2187, opened 2023-04-17
add save flags

Stevent-fei:
support save flags `--tmp-dir`:
`sealer save abc:v1 -o my.tar --tmp-dir /root/my-tmp`


kakaZhou719 (approved):
LGTM

Issue #2147, opened 2023-03-28
Update containers vendor


kakaZhou719:
## Issue Description

1.  support set load tmp dir for save and load, we could set `tmp-dir` arg to `sysContext.BigFilesTemporaryDir`.
2. Waiting for a separate functions from containers SDK, which will not guess what type of input tar file. this is more time-saving.

more to see : https://github.com/containers/common/issues/1376

Type: *feature request*

kakaZhou719:
related go mod version 
```yaml
 github.com/containers/buildah v1.29.1
  github.com/containers/common v0.51.0
  github.com/containers/image/v5 v5.24.1
  github.com/containers/ocicrypt v1.1.7
  github.com/containers/storage v1.45.3
```

cubxxw:
+ [x]  add a flag

  ```bash
  	// TODO: Waiting for implementation, not yet supported
  	flags.StringVar(&loadOpts.TmpDir, "tmp-dir", "", "Set temporary directory when load image. if not set, use system temporary directory(`/var/tmp/`)")  
  ```

+ [ ] come true `MultiImageArchive`  
  ```bash
  	// MultiImageArchive - if true, the container engine allows for storing
  	// archives (e.g., of the docker-archive transport) with multiple
  	// images.  By default, Podman creates single-image archives.
  	MultiImageArchive bool `toml:"multi_image_archive,omitempty"`
  ```

+ [ ] update pack
> Do I need to update the following packages?

+ [ ] Temporary load directory primitives, garbage collection

PR #2193, opened 2023-04-20
feature: Extend sealer save to support multi-image-archive and tmp-dir

cubxxw:
Signed-off-by: Xinwei Xiong(cubxxw) <3293172751nss@gmail.com>


kakaZhou719:
@cubxxw ，how can we support to set tmp dir for load and save cmd. i assume we need below 3 steps:
1. add flag 
2. use it to set sealer tmp dir
3. use it to set buildah tmp dir

could pls recheck our code to implement it?
more to see: https://github.com/sealerio/sealer/issues/2147

Issue #2194, opened 2023-04-20
Use 'uname -n' instead of 'hostname'


Daz-3ux:
## Issue Description

Type: *feature request*

### Describe what feature you want
When setting up a node on the Arch Linux virtual machine, I encountered a problem where the 'hostname' command was not found. After running 'sudo pacman -S inetutils', everything worked fine. Perhaps we can use 'uname -n' as a replacement for 'hostname' to achieve higher universality.

### Additional context
On the next reboot, systemd-hostname will read the /etc/hostname and set the hostname. So `uname -n` seems have no difference with hostname(besides more general).

I can handle this issue if it makes sense : )

Daz-3ux:
error agian👀
```
masterdaz# sealer run kubernetes:v1.19.8  \
  --masters 192.168.122.8 \
  --nodes 192.168.122.9,192.168.122.10 --passwd xx
2023-04-22 14:53:46 [INFO] [local.go:288] Start to create a new cluster: master [192.168.122.8], worker [192.168.122.9 1
92.168.122.10]
copying files to 192.168.122.10: 32/32
copying files to 192.168.122.9: 32/32
Error: failed to init master0 exec remote cmd failed, hostname exit status 127
```

kakaZhou719:
@Daz-3ux , nice issue,  it sounds good to me.  we should get the hostname value from where it was registered.  if `uname -n` meet our request , feel free pr your code 😀

PR #2197, opened 2023-04-23
feature: use 'uname -n' instead of 'hostname' to get hostname

Daz-3ux:
fixes: #2194

The `hostname` command may need to be downloaded on some machines, but the `uname` command is always present.

Fixes: #2194


pkg/infradriver/ssh_infradriver.go
```diff
@@ -306,7 +306,7 @@ func (d *SSHInfraDriver) GetClusterLaunchApps() []string {
 }
 
 func (d *SSHInfraDriver) GetHostName(hostIP net.IP) (string, error) {
-	hostName, err := d.CmdToString(hostIP, nil, "hostname", "")
+	hostName, err := d.CmdToString(hostIP, nil, "uname", "-n")
```
kakaZhou719, line 309: d.CmdToString(hostIP, nil, "uname -n", "")

Daz-3ux: My bad. I've fixed it.

kakaZhou719 (approved):
/lgtm

Issue #2216, opened 2023-05-11
[help wanted] log ssh cmd errors


kakaZhou719:
## Issue Description

Type: *feature request*

### Describe what feature you want

Sealer uses ssh cmd to execute some Linux commands,  but we do not know the real error output message
For example, in the following error, we do not know the real cause.

![image](https://github.com/sealerio/sealer/assets/83740799/589daa59-6f60-4e5f-a82b-528d4d018b41)

so we can log ssh cmd error to make it easier to know what went wrong.

Daz-3ux:
I want to do this, can you assign me : )

kakaZhou719:
> I want to do this, can you assign me : )

hhh, feel free to pr your code.

PR #2217, opened 2023-05-12
feat: log ssh exit error

Daz-3ux:
tell the user that the error reason is ssh session exit meet error, 
not just display the error number.

Fixes: #2216

### DIFF
--- a/utils/ssh/sshcmd.go
+++ b/utils/ssh/sshcmd.go
@@ -71,7 +71,7 @@ func (s *SSH) CmdAsync(host net.IP, hostEnv map[string]string, cmds ...string) e
 				return fmt.Errorf("failed to start command %s: %v", cmd, err)
 			}
 
-			_ = ReadPipe(stdout, stderr, s.AlsoToStdout)
+			ReadPipe(stdout, stderr, s.AlsoToStdout)
 
 			err = c.Wait()
 			if err != nil {
@@ -100,11 +100,11 @@ func (s *SSH) CmdAsync(host net.IP, hostEnv map[string]string, cmds ...string) e
 				return fmt.Errorf("failed to start command %s on %s: %v", cmd, host, err)
 			}
 
-			errMsg := ReadPipe(stdout, stderr, s.AlsoToStdout)
+			ReadPipe(stdout, stderr, s.AlsoToStdout)
 
 			err = session.Wait()
 			if err != nil {
-				return fmt.Errorf("failed to execute command(%s) on host(%s): error(%v) and errMsg(%v)", cmd, host, err, errMsg)
+				return fmt.Errorf("failed to execute command(%s) on host(%s): error(%v)", cmd, host, err)
 			}
 
 			return nil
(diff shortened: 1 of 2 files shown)


kakaZhou719:
@Daz-3ux ，we can not modify vendor files directly unless we update go mod. BTW, try to fix it at those code: https://github.com/sealerio/sealer/blob/150c4cbaba9ce0942141c6f30c14d7cf7821521d/utils/ssh/sshcmd.go#L46

Daz-3ux:
> @Daz-3ux ，we can not modify vendor files directly unless we update go mod. BTW, try to fix it at those code:
> 
> https://github.com/sealerio/sealer/blob/150c4cbaba9ce0942141c6f30c14d7cf7821521d/utils/ssh/sshcmd.go#L46

Of course, I will fix it

Daz-3ux:
@kakaZhou719 
I commit a new, but now stderr is being passed twice. Please check if it's feasible to log the error in this way

Daz-3ux:
> 

It can print error message now and not just error code
image

utils/ssh/sshcmd.go
```diff
@@ -100,11 +100,11 @@ func (s *SSH) CmdAsync(host net.IP, hostEnv map[string]string, cmds ...string) e
 				return fmt.Errorf("failed to start command %s on %s: %v", cmd, host, err)
 			}
 
-			ReadPipe(stdout, stderr, s.AlsoToStdout)
+			errMsg := ReadPipe(stdout, stderr, s.AlsoToStdout)
```
kakaZhou719, line 103: how do we read message from a chan ? i think it is no need to return it, just print it to stdout

utils/ssh/utils.go
```diff
@@ -59,7 +59,7 @@ func fileExist(path string) bool {
 	return err == nil || os.IsExist(err)
 }
 
-func ReadPipe(stdout, stderr io.Reader, alsoToStdout bool) {
+func ReadPipe(stdout, stderr io.Reader, alsoToStdout bool) error {
 	var combineSlice []string
```
kakaZhou719, line 63: combineSlice will contains all the cmd output. so we do new combineErrorSlice to gather  error message .

utils/ssh/sshcmd.go
Daz-3ux: thx for your review, I will fix my code

utils/ssh/utils.go
```diff
@@ -60,18 +60,21 @@ func fileExist(path string) bool {
 }
 
 func ReadPipe(stdout, stderr io.Reader, alsoToStdout bool) {
-	var combineSlice []string
-	var combineLock sync.Mutex
+	var combineDoneSlice []string
+	var combineDoneLock sync.Mutex
+	var combineErrSlice []string
```
kakaZhou719, line 65: good , if we want to see specific error, `combineErrSlice` will contain all messages.

Daz-3ux: Maybe we can pass a flag and filter to readPipe(), then we can select error
such as:
```
if strings.Contains(message, errorFilter) {
			fmt.Println("Specific error:", message)
		}
```

kakaZhou719: yes, agree with you.@Daz-3ux

PR #2218, opened 2023-05-15
Feat/make all masters equal and make join cmd more robust

VinceCui:
This PR support the follow abilities:
- Support clean of an unsuccessfully installed master/node
- Make all masters equal and NOT allow clean master from itself
- When some node missing or passwd changed, will not block scaling up new nodes
- When some node missing or passwd changed, support scaling down them directly
- Force specify PATH env to avoid executable files that cannot be found in some environments
- Improve some error lint.

### DIFF
--- a/pkg/runtime/kubernetes/init.go
+++ b/pkg/runtime/kubernetes/init.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


pkg/runtime/kubernetes/init.go
```diff
@@ -237,9 +241,22 @@ func (k *Runtime) initKube(hosts []net.IP) error {
 
 func (k *Runtime) sendClusterCert(hosts []net.IP) error {
 	f := func(host net.IP) error {
-		err := k.infra.Copy(host, k.getPKIPath(), clustercert.KubeDefaultCertPath)
-		if err != nil {
-			return fmt.Errorf("failed to copy cluster cert : %v", err)
+		if err := k.infra.Copy(host, k.getPKIPath(), clustercert.KubeDefaultCertPath); err != nil {
+			return fmt.Errorf("failed to copy cluster cert: %v", err)
+		}
+		if err := k.infra.Copy(host, k.getPKIPath(), k.getPKIPath()); err != nil {
```
Starnop, line 247: What does this mean?

VinceCui: transform pki to all master's rootfs, so when master0 down, other master can scale up cluster.

Starnop (approved):
LGTM

Issue #2226, opened 2023-05-18
无法删除feature-gates参数


fangzhengjin:
### What happen?

在`kubeadm.yml.tmpl`中删除`feature-gates`字段或将`feature-gates`字段设置为空字符串, sealer都会覆盖该内容为`TTLAfterFinished=true,EphemeralContainers=true`

但是在高版本设置了`TTLAfterFinished=true,EphemeralContainers=true`会导致集群无法初始化

### What is the version of Sealer you using?

{"gitVersion":"v0.10.0","gitCommit":"d83ead0","buildDate":"2023-05-16 02:39:03","goVersion":"go1.17.13","compiler":"gc","platform":"linux/amd64"}

### What is your OS environment?

Kylin Linux Advanced Server V10 (Sword)

### What is the Kernel version?

Linux kubemtr01 4.19.90-24.4.v2101.ky10.x86_64 #1 SMP Mon May 24 12:14:55 CST 2021 x86_64 x86_64 x86_64 GNU/Linux

### Other environment you want to tell us?

- Cloud provider or hardware configuration:
- Install tools:
- Others:

fangzhengjin:
同时还发现了在k8s v1.19 开始 `--experimental-cluster-signing-duration`参数也作废了, 会导致`kube-controller-manager`无法启动

fangzhengjin:
sealer能否提供不合并kubeadm.yml的参数配置? 不然有些参数删又删不了 改也改不了 还会导致集群无法启动

fangzhengjin:
如果一定要合并的话不如参考下kubekey这种, 先校验再看要不要合入

https://github.com/kubesphere/kubekey/blob/4bc171e0088be251ee492d92701d0825bedda480/cmd/kk/pkg/kubernetes/templates/v1beta2/kubeadm_config.go#L215-L263
<img width="1472" alt="image" src="https://github.com/sealerio/sealer/assets/12680972/c1d270f9-898d-49d2-8a25-2c152d7884d8">

kakaZhou719:
@fangzhengjin ,这里的问题应该是和sealer代码中内置的kubeadm 配置文件合并导致的。我觉得默认的kubeadm.yaml, 不应该在sealer代码中，而是应该和sealer 镜像保持同步，这样的话，就能避免这个版本问题了。

PR #2229, opened 2023-05-19
feat: add featureGates optimization

fangzhengjin:
When merging the `kubeadm.yml` file, remove `featureGates` that have been removed in the current version of kubernetes to prevent cluster initialization failures.
Like this:
image

Link: https://github.com/sealerio/sealer/issues/2226


kakaZhou719:
@fangzhengjin , i think it’s better to bound `kubeadm.yaml` to the specific image version, not included in our code.

fangzhengjin:
> @fangzhengjin , i think it’s better to bound `kubeadm.yaml` to the specific image version, not included in our code.

I think this function makes sense. Regardless of whether part of the `kubeadm.yaml` is in the `sealer` code or not, it can delete parameters that should not be used according to the invalid parameters listed by Kubernetes to prevent the cluster from being able to start.

PR #2233, opened 2023-05-23
feat: save application spec to cluster

kakaZhou719:

1.  save application spec to cluster
2.  add log message


starnop (approved):
LGTM

PR #2234, opened 2023-05-23
optimize: cluster cmd

kakaZhou719:

Nothing to do with logic changes. below list is what this PR does:

1. add some log message
2. aggregated  cmd example and description together
3. new two object `kubeInstaller` and `appInstaller`. one for cluster management and another one is for app management
4. each subcommand only do parameter judgments, put the real operation logic into the installer.


cmd/sealer/cmd/cluster/cmd.go
```diff
@@ -0,0 +1,100 @@
...
+	applyFlags    = &types.ApplyFlags{}
+	scaleUpFlags  = &types.ScaleUpFlags{}
+	joinFlags     = &types.ScaleUpFlags{}
+	rollbackFlags = &types.RollbackFlags{}
+	upgradeFlags  = &types.UpgradeFlags{}
+	deleteFlags   = &types.DeleteFlags{}
+)
+
+// long cmd cli message
+var (
```
Starnop, line 34: In fact, I personally think it might be better for each cmd to maintain all of its own information so that fewer modifications are involved when we add or deprecated a command

kakaZhou719: yes,  we do better keep each description with its own cmd. but if we have much more subcommand , aggregated them together may more easier to maintain.

starnop: I don't think so, unless you have a need to batch modify subcommands

starnop (approved):
LGTM

PR #2238, opened 2023-06-01
bugfix: add lock to image blob list

kakaZhou719:

bugfix: add lock to image blob list, slice is not concurrency safe 

image


starnop (approved):
LGTM

PR #2246, opened 2023-06-14
opt: sealer common values

kakaZhou719:

no logic changed, just remove some unused code to make code more clearly


pkg/logger/file_hook.go
```diff
@@ -29,7 +29,7 @@ import (
 
 func NewFileHook(filePath string) (logrus.Hook, error) {
 	if filePath == "" {
-		filePath = common.DefaultLogDir
+		filePath = common.GetSealerWorkDir()
```
starnop, line 32: the log dir will be changed here?

PR #2249, opened 2023-06-15
bugfix: return error if scale ip already in cluster

kakaZhou719:

1. bugfix: return error if scale ip already in cluster
2. optimize the delete prompt


starnop (approved):
LGTM

PR #2251, opened 2023-06-21
feat: sealer build supports lite mode option

kakaZhou719:

In some case, some applications do not need to store additional container images. it is faster and smaller.

so, at build stage, sealer could try to skip download container image through build mode option.

if it is `lite`,  will skip download container image. default value is `all`, means download all container image.

how to use: 

```
sealer build -t abc:v1 --build-mode=lite
sealer build -t abc:v1 --build-mode=lite --platform linux/arm64
sealer build -t abc:v1 --build-mode=lite --platform linux/amd64,linux/arm64
```


starnop (approved):
LGTM

PR #2254, opened 2023-06-26
bugfix: add platform value to container image list 

kakaZhou719:
1. add platform to container image list;
2. skip download duplicate container image;
3. add some debug logs

### DIFF
--- a/pkg/image/save/save.go
+++ b/pkg/image/save/save.go
@@ -71,6 +71,7 @@ func (is *DefaultImageSaver) SaveImages(images []string, dir string, platform v1
 		}
 	}()
 
+	existFlag := make(map[string]struct{})
 	//handle image name
 	for _, image := range images {
 		named, err := ParseNormalizedNamed(image, "")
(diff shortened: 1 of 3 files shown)


pkg/image/save/save.go
```diff
@@ -71,14 +71,22 @@ func (is *DefaultImageSaver) SaveImages(images []string, dir string, platform v1
...
+	existFlag := make(map[string]struct{})
 	//handle image name
 	for _, image := range images {
 		named, err := ParseNormalizedNamed(image, "")
 		if err != nil {
 			return fmt.Errorf("failed to parse image name:: %v", err)
 		}
 
-		//check if image exist
+		//check if image is duplicate
```
starnop, line 82: This judgment may be redundant, as the OCI pull process itself checks for the presence of the image layer.

In addition, there is a risk of misjudgment here.

kakaZhou719: 1. i wonder there is not redundant.
for example: input images["nginx:latest","docker.io/library/nginx:latest"], will be transformed ["docker.io/library/nginx:latest","docker.io/library/nginx:latest"], so we need to do deduplication，nothing to do with pull process logic.

2. what risks here ? sorry i dont understand?

starnop: Sorry, I thought we were doing the duplicate removal in the step of pulling image. Currently, it seems that only for the image name, which is OK

starnop (approved):
LGTM

PR #2255, opened 2023-06-28
 refactor: set appcmds value to application

kakaZhou719:

1. set appcmds value to imageExtension.application 
2. refactor new app driver code, format it firstly and use application spec to override it.
3. add some UT test


starnop (approved):
LGTM

PR #2262, opened 2023-07-12
feat: add pre check for cluster install

kakaZhou719:
feat: add pre check for cluster install

### DIFF
--- a/pkg/preflight/checker.go
+++ b/pkg/preflight/checker.go
@@ -1,0 +1,46 @@
+// Copyright © 2023 Alibaba Group Holding Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


pkg/preflight/checker.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Name returns label for checker.
+func (ClusterIsExistsCheck) Name() string {
+	return "ClusterExist"
+}
+
+// Check :if the given cluster is existed, return error.
+func (c ClusterIsExistsCheck) Check() (warnings, errorList []error) {
+	// check K8s client
+	client, _ := k8s.NewK8sClient()
```
starnop, line 40: I wonder if this method will take too long time and there will be misjudgments when the API server is abnormal

kakaZhou719: yes, the granularity is too large.and there will be a lot of unpredictable things to new k8s client.

However , i think we take master0 as sealer executor machine, if sealer could new k8s client on master0 , master0 must exist in a cluster, so we need to return  error.

Correct me if i am wrong.

PR #2311, opened 2023-10-25
chore: fix golang ci lint error

starnop:

1. ignore `has no field or method` error from `typecheck` linter
2. replace deprecated linter to suggested replaced linter
3. fix gosec error


VinceCui (approved):
LGTM

PR #2312, opened 2023-10-26
Bumped the version of golangci-lint

dynos01:

- Bumped the version of `golangci-lint` to `1.44.0` (this version uses far less memory than previous versions)
- Fixed one line of error message that cannot pass new lints


VinceCui (commented):
LGTM

VinceCui (approved):

